PR #5, opened 2020-06-07
Add startup errors, mention XDG_RUNTIME_DIR in README

arp242:
It was just exiting with code 1 for me. The problem turned out to be
that you *need* to set XDG_RUNTIME_DIR or it won't work (I think e.g.
systemd does that for you, but on Void it's not set by default), so
mention that in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,10 @@ this process a SIGTERM and wait for it to terminate (if it hasn't already).
 This makes it ideal not only for initialization but also for execing into a
 user-level service manager like s6 or `systemd --user`.
 
+You'll have to explicitly set `XDG_RUNTIME_DIR` if your system doesn't do it for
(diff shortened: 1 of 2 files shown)


djpohly:
Ah, Wayland puts its communication socket in `$XDG_RUNTIME_DIR`, so that needs to be set, and if you're not using elogind (or alternatives) then it might not be configured.  I've added a bit of code to check explicitly for this variable as well as including the note on the README and the missing `perror`s you found.  Thanks! :)

arp242:
Yeah, cheers. You get an error message about it from Wayland if you run it with `-v`, but not a bad thing to be explicit about that. 

Seems a bit strange it doesn't show any message to that by default as the default loglevel is `WLR_ERROR`, and this seems like an error to me (similar situation if the directory doesn't exist: `-v` gives errors, but not with `WLR_ERROR`)

This is my first experience with Wayland, so maybe it fits with how it's designed to work.

PR #8, opened 2020-06-13
Make sure to not pass NULL surfaces to these wlr functions

Shugyousha:
On my machine passing NULL surfaces resulted in aborts.

I am not sure this is the right way to fix these so you should have a closer look at the issue.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -907,7 +907,7 @@ pointerfocus(Client *c, struct wlr_surface *surface, double sx, double sy,
 
 	wlr_seat_pointer_notify_enter(seat, surface, sx, sy);
 	/* If keyboard focus follows mouse, enforce that */
-	if (sloppyfocus && surface)
+	if (sloppyfocus)
 		focusclient(c, surface, 0);
 }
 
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -897,8 +897,14 @@ pointerfocus(Client *c, struct wlr_surface *surface, double sx, double sy,
...
 	/* If surface is NULL, clear pointer focus, otherwise let the client
 	 * know that the mouse cursor has entered one of its surfaces. */
+	if (!surface) {
+		wlr_seat_pointer_notify_clear_focus(seat);
+		return;
+	}
+
 	wlr_seat_pointer_notify_enter(seat, surface, sx, sy);
 	/* If keyboard focus follows mouse, enforce that */
 	if (sloppyfocus && surface)
```
djpohly, line 910:
```suggestion
	if (sloppyfocus)
```

dwl.c
```diff
@@ -597,7 +597,7 @@ focusclient(Client *c, struct wlr_surface *surface, int lift)
 	 * automatically send key events to the appropriate clients.  If surface
 	 * is NULL, this will clear focus.
 	 */
-	if (surface != psurface) {
+	if (surface && (surface != psurface)) {
```
djpohly, line 600: I think this needs a corresponding clear_focus as well.  How about `if (!surface)` and then `else if (surface != psurface)`?

dwl.c
Shugyousha: Makes sense!

djpohly:
Thanks!

PR #13, opened 2020-06-30
Add repeat delay and repeat rate to config.h

gustafla:
Hi and thanks for this cool wm. After switching to sway from dwm, I've been missing something exactly like this (and also a good wayland-port of st). I think it's better to have keyboard repeat rate and -delay available in the config.h instead of hardcoding them in dwl.c, because I really need to change them to feel comfortable on my computer. This PR addresses that.


djpohly:
Looks great.  Thanks for the contribution!

Issue #19, opened 2020-07-25
Atomic layout updates (eliminate flicker on window resize)


djpohly:
When the window layout changes, dwl should not render a frame with the new layout until all of the visible windows can be drawn at their new sizes.

Currently it is possible to draw frames in which the size/position/visibility of windows are in an intermediate state.  For example, when a new tiled window is mapped, the other tiled windows may commit their new-sized buffer and be redrawn in their updated positions before the new window appears, showing the background through a "hole," or a window which needs to be sized down will be drawn in its new position before its buffer changes size, overlapping with other windows.

The solution doesn't seem to be short and obvious, so if there is not a simple and fully correct way to do this, then let's aim for simple and sufficient before complex and fully correct.

alex-courtis:
A quick debug has confirmed that the same thread is invoking rendermon as invokes tile. Therefore we are not looking at a simple problem of concurrent data access.

Confirming my understanding of the example above: A new surface is created, and is not ready to be rendered at createnotify time, but is ready once maprequest is invoked.

Solutions for that example might include:
1. Block rednermon between createnotify and maprequest
2. Do not include the new surface in the client list until maprequest
3. Block arrange functions between createnotify and maprequest

The general solution involves identifing pairs of signal functions that demarcate these intermeditate states and avoiding display of those states.

Please advise whether this is a correct summary of the problem, or if there is something bigger that I am not seeing.

djpohly:
Correct, this isn't about concurrent data access.  Since Wayland is asynchronous, there's time (and potentially frame-render requests) during the "configure - ack_configure - commit" cycle for each window that is being created or resized, and that time may differ for each window.  `arrange` sends the initial configure to all of the windows, but we don't want to draw *any* window at its new size/position until *all* of the windows have committed at their new sizes (or, at least, until we can draw something reasonable at their new sizes).

alex-courtis:
Understood.

Sway has a transaction system that blocks render during transactions, although that appears to be more to handle sway's arrangement of windows being spread across many different places.

Transactions are a very heavy solution; something simpler such as dirty client state might suffice for dwl.

djpohly:
I suspect we'll have to add a surface.commit handler and do something with `configure_serial`.  We don't have a buffer at the new size until the first commit after ack_configure, and it doesn't look like wlroots does anything special at commit time to record which xdg_shell state changes have actually been acknowledged.  swaywm/wlroots#762 and swaywm/wlroots#1657 both sound related.

[...]

PR #20, opened 2020-07-26
an attempt at atomic arrange

djpohly:
Currently freezes rendering with clients that ignore our size demands,
but it seems to eliminate resize flicker for well-behaved programs.

An idea for #19.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1214,6 +1214,8 @@ void
 rendermon(struct wl_listener *listener, void *data)
 {
 	struct wlr_output *output = data;
+	Client *c;
+	int defer = 0;
 
 	/* This function is called every time an output is ready to display a frame,
 	 * generally at the output's refresh rate (e.g. 60Hz). */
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -1222,6 +1224,21 @@ rendermon(struct wl_listener *listener, void *data)
 	struct timespec now;
 	clock_gettime(CLOCK_MONOTONIC, &now);
 
+	wl_list_for_each(c, &stack, slink) {
+		if (c->xdg_surface->role != WLR_XDG_SURFACE_ROLE_TOPLEVEL)
+			continue;
+		if (c->xdg_surface->toplevel->server_pending.width !=
+				c->xdg_surface->surface->current.width ||
+				c->xdg_surface->toplevel->server_pending.height !=
+				c->xdg_surface->surface->current.height) {
```
alex-courtis, line 1233: Nice idea; shame we don't have position info in server_pending.

djpohly: Nah, position info isn't needed here, since Wayland clients are position-agnostic.  Resizing nicely, however, requires a round-trip to the client.

alex-courtis: Good to know; thanks!

PR #25, opened 2020-07-27
getatom: fix potential uninitialized atom variable

geistesk:
The getatom function returns the atom variable, which is only
initialized in case of a success. This results in a maybe-uninitialized
warning/error. After this commit, now a zero value is returned in case
of error.

~~Furthermore, the getatom calls in the xwaylandready function are now
checked against zero atoms.~~
_(Edit: This was removed; see conversation below)_

Previously, the compilation errors with:
```
dwl.c: In function ‘getatom’:
dwl.c:776:9: error: ‘atom’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
  776 |  return atom;
      |         ^~~~
cc1: all warnings being treated as errors
make: *** [<builtin>: dwl.o] Error 1
```

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -762,7 +762,7 @@ focusstack(const Arg *arg)
 Atom
 getatom(xcb_connection_t *xc, const char *name)
 {
-	Atom atom;
+	Atom atom = 0;
 	xcb_generic_error_t *error;
 	xcb_intern_atom_cookie_t cookie;
 	xcb_intern_atom_reply_t *reply;


dwl.c
```diff
@@ -1724,6 +1728,13 @@ xwaylandready(struct wl_listener *listener, void *data) {
 	netatom[NetWMWindowTypeUtility] = getatom(xc, "_NET_WM_WINDOW_TYPE_TOOLBAR");
 	netatom[NetWMWindowTypeToolbar] = getatom(xc, "_NET_WM_WINDOW_TYPE_UTILITY");
 
+	for (i = 0; i < NetLast; i++) {
+		if (!netatom[i]) {
+			fprintf(stderr, "getatom failed for netatom %ld", i);
+			return;
+		}
+	}
+
```
djpohly, line 1737: Do we really need this?  Can we still proceed if one of the atoms doesn't exist?

alex-courtis: We can live without atoms; we just won't have the functionality they indicate.

See https://github.com/djpohly/dwl/blob/8b7cc8ada65939b12320b9c77d656401c1ca9f15/dwl.c#L1720

dwl.c
```diff
@@ -771,6 +771,8 @@ getatom(xcb_connection_t *xc, const char *name)
 	reply = xcb_intern_atom_reply(xc, cookie, &error);
 	if (reply != NULL && error == NULL)
 		atom = reply->atom;
+	else
+		atom = 0;
```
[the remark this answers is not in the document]
geistesk: Changed

dwl.c
geistesk: Okay, I have removed those checks. I just felt _unsafe_ without checking for potential error values.

geistesk:
Thanks for your review. I have updated the PR regarding your comments. Please take another look.

djpohly:
It's good programming discipline to consider the return values, I agree, and the code should reflect that.  In this case, since the program didn't actually need to do anything, I added to the comment where `getatom` is called to explain why ignoring the error return is the correct behavior.

Thanks for pointing this out!

PR #27, opened 2020-07-28
Allow toggling the layout before selecting a different one

guidocella:


djpohly:
I don't think I'm feeling this one for mainline dwl.  If you haven't changed the layout yet, I think it makes sense that there's not a previous layout.  Plus it adds a non-obvious line of code for a case that will happen at most once per monitor, and it's easy enough to switch the layout explicitly the first time.

guidocella:
When it creates a monitor dwm explicitly sets the previous layout to the second one to allow this
```c
m->lt[0] = &layouts[0];
m->lt[1] = &layouts[1 % LENGTH(layouts)];
```
I personally find this convenient and only switch layouts by toggling between tile and monocle (which I made 2nd in layouts array), without having to remember if I already switched, and never use the key bindings to select a specific layout.

The line can be made clearer by making it longer, for example
```c
m->lt[1] = LENGTH(layouts) > 1 && r->lt != &layouts[1] ? &layouts[1] : &layouts[0];
```
or just assume the common case that the monitor's first layout is &layouts[0] and do like dwm
```c
m->lt[1] = &layouts[1 % LENGTH(layouts)];
```

Anyway, your call.

alex-courtis:
This is a valid use case for me: I expect dwm to toggle between the first two layouts (monocle, tile) when it first starts up.

Further to that, I'm using the [pertag](http://dwm.suckless.org/patches/pertag/) patch, and don't want to have to manually setup the toggle state for every tag.

djpohly:
Hm.  It's not the job of the core codebase to support an external patch like pertag - that responsibility belongs to the patch.  On the other hand, parity with dwm is a legitimate point.  dwm doesn't support setting the initial layout like dwl, and I'm trying to discern what is the least surprising behavior given that difference.

I'll leave the issue open for now since I don't feel it's 100% resolved, but I went ahead and linked this patch on the new [user-contributed patches page](https://github.com/djpohly/dwl/wiki/Patches).  If you don't want it there, you can feel free to take it down.

MadcowOG:
I think this is a valid pull request, dwm does this. If the goal is for dwl to have feature parity with dwm, then this should be merged.

PR #28, opened 2020-07-29
do not render frames when any XDG surface has uncommitted changes

alex-courtis:
~~This is not a complete solution. We can use this mechanism to skip more frames as we discover means of detecting them.~~

~~The commit handler is picking up the uncommitted changes when creating a new client, however does not see other changes such as a master width resize.~~

Update: This change is complete, for XDG clients only. Resizes are aggressively marked as dirty, and skip rendering until all the changes have been committed.

Such a solution cannot work for X.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -75,6 +75,7 @@ typedef struct {
 		struct wlr_xwayland_surface *xwayland_surface;
 	};
 	struct wl_listener activate;
+	struct wl_listener commit;
 	struct wl_listener map;
 	struct wl_listener unmap;
 	struct wl_listener destroy;
@@ -84,6 +85,7 @@ typedef struct {
 	int bw;
 	unsigned int tags;
 	int isfloating;
+	int dirty;
 } Client;
 
 typedef struct {
@@ -443,6 +445,15 @@ cleanupmon(struct wl_listener *listener, void *data)
 }
 
 void
+commitnotify(struct wl_listener *listener, void *data)
+{
+	Client *c = wl_container_of(listener, c, commit);
+
+	/* client is dirty if there are any changes that have not been committed */
+	c->dirty = c->xdg_surface->configure_serial != c->xdg_surface->configure_next_serial;
+}
+
+void
 createkeyboard(struct wlr_input_device *device)
 {
 	struct xkb_context *context;
@@ -553,6 +564,8 @@ createnotify(struct wl_listener *listener, void *data)
 			WLR_EDGE_BOTTOM | WLR_EDGE_LEFT | WLR_EDGE_RIGHT);
 
 	/* Listen to the various events it can emit */
+	c->commit.notify = commitnotify;
+	wl_signal_add(&xdg_surface->surface->events.commit, &c->commit);
 	c->map.notify = maprequest;
 	wl_signal_add(&xdg_surface->events.map, &c->map);
 	c->unmap.notify = unmapnotify;
@@ -625,6 +638,7 @@ destroynotify(struct wl_listener *listener, void *data)
 {
 	/* Called when the surface is destroyed and should never be shown again. */
 	Client *c = wl_container_of(listener, c, destroy);
+	wl_list_remove(&c->commit.link);
 	wl_list_remove(&c->map.link);
 	wl_list_remove(&c->unmap.link);
 	wl_list_remove(&c->destroy.link);
@@ -1218,6 +1232,8 @@ void
 rendermon(struct wl_listener *listener, void *data)
 {
 	struct wlr_output *output = data;
+	Client *c;
+	int render = 1;
 
 	/* This function is called every time an output is ready to display a frame,
 	 * generally at the output's refresh rate (e.g. 60Hz). */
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -1226,16 +1242,29 @@ rendermon(struct wl_listener *listener, void *data)
...
-	wlr_renderer_clear(drw, rootcolor);
 
-	renderclients(m, &now);
-	renderindependents(output, &now);
+	if (render) {
+		wlr_renderer_clear(drw, rootcolor);
+
+		renderclients(m, &now);
+		renderindependents(output, &now);
+	}
```
alex-courtis, line 1267: It appears necessary to call render begin / end, otherwise the loop calling render seems to... stop.

dwl.c
```diff
@@ -442,6 +444,15 @@ cleanupmon(struct wl_listener *listener, void *data)
 	free(m);
 }
 
+void
+commitnotify(struct wl_listener *listener, void *data)
+{
+	Client *c = wl_container_of(listener, c, commit);
+
+	/* client is dirty if there are any changes that have not been committed */
+	c->dirty = c->xdg_surface->configure_serial != c->xdg_surface->configure_next_serial;
```
alex-courtis, line 453: Whoops... this doesn't cover xwayland.

Fixing...

alex-courtis: Configure serial is XDG only, hence we can talk about xdg_surface.

It kind of makes sense, as frame by frame updates of applications is not really something X11 really does.

dwl.c
```diff
@@ -1264,6 +1295,7 @@ resize(Client *c, int x, int y, int w, int h, int interact)
 	c->geom.y = y;
 	c->geom.width = w;
 	c->geom.height = h;
+	c->dirty = 1;
```
alex-courtis, line 1298: The dwm/dwl structure makes our lives easy: everything happens in resize.

dwl.c
```diff
@@ -84,6 +85,7 @@ typedef struct {
 	int bw;
 	unsigned int tags;
 	int isfloating;
+	int dirty;
```
djpohly, line 88: Maybe `resizing` instead - "dirty" could mean a lot more than just that.

dwl.c
djpohly, line 453: This would pause rendering for any configure-able property, and not just resizes.  Don't set the flag here, only clear it:
```suggestion
	c->dirty = c->dirty && c->xdg_surface->configure_serial != c->xdg_surface->configure_next_serial;
```
or
```suggestion
	if (c->xdg_surface->configure_serial != c->xdg_surface->configure_next_serial)
		c->dirty = 0;
```
(`&=` would work but is a little too tricky for my taste, and `&&=`, sadly, doesn't exist)

dwl.c
djpohly: Hmm. The frame_done events made it work for me, but I'd really like to know where the hangup is so I know we're doing the right thing here.

(Also, brace style on the `rendermon` changes doesn't match the rest of the code.)

djpohly:
Yeah, X doesn't have the mechanism to make this work.  It's legacy anyway, so I'm not going to waste time/lines trying to X clients pretty.

This looks sufficient.  It might be able to be tightened up a bit later using the serials returned from set_size, or by filtering which clients can actually delay a render, but I think it's good for now.

dwl.c
alex-courtis: Changed to resize, which is the configure serial of the outstanding operation.

dwl.c
alex-courtis: Specifically targets the configure serial of the resize now.

alex-courtis:
> This looks sufficient. It might be able to be tightened up a bit later using the serials returned from set_size

That worked beautifully - fantastic idea.

dwl.c
alex-courtis: > Hmm. The frame_done events made it work for me, but I'd really like to know where the hangup is so I know we're doing the right thing here.

I'm not sure it was actually working: I tried the atomic-arrange branch, adding some log lines to rendermon. It stops being called the first time we set defer = 1.

Digging into wlroots / sway to see what is going on.

alex-courtis: sway has set up its damage detection in such a way that its render callback `damage_handle_frame` is only invoked when there are changes to render. `wlr_renderer_begin` / `end` are not called during that callback; instead they are called as part of sway's transaction handling mechanism, as a result of user input. This explains nothing.

wlroots invokes rendermon in response to wayland DRM events:
```
#0  rendermon (listener=0x555555f1a888, data=0x555555f19f40) at dwl.c:1248
#1  0x00007ffff7f7fc08 in wlr_signal_emit_safe (signal=0x555555f1a0c8, data=0x555555f19f40) at ../util/signal.c:29
#2  0x00007ffff7f6da6a in wlr_output_send_frame (output=0x555555f19f40) at ../types/wlr_output.c:661
#3  0x00007ffff7f28452 in page_flip_handler (fd=9, seq=689548, tv_sec=11553, tv_usec=356413, crtc_id=51, data=0x55555558b2a0)
    at ../backend/drm/drm.c:1527
#4  0x00007ffff7c47477 in drmHandleEvent () at /usr/lib/libdrm.so.2
#5  0x00007ffff7f284d7 in handle_drm_event (fd=9, mask=1, data=0x0) at ../backend/drm/drm.c:1537
#6  0x00007ffff7ec8faa in wl_event_loop_dispatch () at /usr/lib/libwayland-server.so.0
#7  0x00007ffff7ec74e7 in wl_display_run () at /usr/lib/libwayland-server.so.0
#8  0x000055555555c729 in run (startup_cmd=0x0) at dwl.c:1371
#9  0x000055555555db00 in main (argc=1, argv=0x7fffffffe838) at dwl.c:1874
```
An apparent side effect of not calling `wlr_renderer_begin` / `end` is these events not being generated or being filtered out.

I have not answered anything here, however this should all become clear as we implement damage detection.

djpohly:
Merged in d6a18d6.  A little hesitant, as it seems that code comprehension is starting to get away from us here... but I really wanted those flickers gone. I would like to make sure we know why each line is there (and document it when not obvious) before taking this any further.

alex-courtis:
> A little hesitant, as it seems that code comprehension is starting to get away from us here... but I really wanted those flickers gone. I would like to make sure we know why each line is there (and document it when not obvious) before taking this any further.

Agreed.

I don't see this code being in the final dwl with damage detection.

Issue #24, opened 2020-07-27
X11 "X" cursor visible when it shouldn't be


djpohly:
When running Firefox in XWayland, the mouse cursor turns into the X11 "X" cursor when a second-level menu such as View->Zoom is open.  Quitting the program also turns the mouse to this cursor until the mouse is moved again.

AFAIK this cursor isn't ever requested by Firefox and so shouldn't be showing up.

(Note: `XC_X_cursor` has the value 0.  Might be relevant, might not.)

alex-courtis:
I reckon we'll need to handle the xwayland cursor notifications and set the appropriate cursor.

alex-courtis:
There are no notifications... setting the default cursor did the job: https://github.com/djpohly/dwl/pull/32/files

pm4rcin:
I guess it could be closed.

PR #32, opened 2020-08-04
#24 ensure that xwayland cursor defaults to left_ptr

alex-courtis:
Set the default cursor for X to left_ptr.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -251,9 +251,11 @@ static struct wl_list independents;
 static struct wlr_xdg_decoration_manager_v1 *xdeco_mgr;
 
 static struct wlr_cursor *cursor;
-static struct wlr_xcursor *xcursor;
 static struct wlr_xcursor_manager *cursor_mgr;
+#ifdef XWAYLAND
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -243,7 +243,9 @@ static struct wl_list independents;
 static struct wlr_xdg_decoration_manager_v1 *xdeco_mgr;
 
 static struct wlr_cursor *cursor;
+static struct wlr_xcursor *xcursor;
 static struct wlr_xcursor_manager *cursor_mgr;
+static struct wlr_xcursor_manager *xcursor_mgr;
```
alex-courtis, line 248: These two could be local.

alex-courtis:
This is ready for merge @djpohly

alex-courtis:
> Hm... why does this require two xcursor_managers?

I took inspiration from sway here: it creates two xcursor_managers. I'd prefer not to share between the wayland/xwayland worlds lest we end up coupling them.

We could try with one and see how it goes...

PR #34, opened 2020-08-04
Keep clients in their own boxes

djpohly:
Clients have no business drawing over top of others - it's a security risk and bad for invariants.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1169,6 +1169,18 @@ renderclients(Monitor *m, struct timespec *now)
 			{ox, oy + c->bw + h, w + 2 * c->bw, c->bw}, /* bottom */
 		};
 
+		/* This calls our render function for each surface among the
+		 * xdg_surface's toplevel and popups. */
+		rdata.output = m->wlr_output;
+		rdata.when = now;
+		rdata.x = c->geom.x + c->bw;
+		rdata.y = c->geom.y + c->bw;
+		wlr_renderer_scissor(drw, &c->geom);
+		if (c->type != XDGShell)
+			wlr_surface_for_each_surface(c->surface.xwayland->surface, render, &rdata);
+		else
+			wlr_xdg_surface_for_each_surface(c->surface.xdg, render, &rdata);
+
 		/* Draw window borders */
 		color = (c == sel) ? focuscolor : bordercolor;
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -1212,8 +1214,19 @@ renderindependents(struct wlr_output *output, struct timespec *now)
 		rdata.x = c->surface.xwayland->x;
 		rdata.y = c->surface.xwayland->y;
 
+		if (c->surface.xwayland->parent) {
+			pg.x = c->surface.xwayland->parent->x;
+			pg.y = c->surface.xwayland->parent->y;
+			pg.width = c->surface.xwayland->parent->width;
+			pg.height = c->surface.xwayland->parent->height;
+			wlr_renderer_scissor(drw, &pg);
```
alex-courtis, line 1222: Firefox and IntelliJ menus follow this branch and are indeed restricted to the parent window.

Chromium menus and IntelliJ windows follow the other branch and are free to render outside the parent.

dwl.c
```diff
@@ -1169,33 +1169,35 @@ renderclients(Monitor *m, struct timespec *now)
...
+	wlr_renderer_scissor(drw, NULL);
 }
 
 void
 renderindependents(struct wlr_output *output, struct timespec *now)
 {
 	Client *c;
 	struct render_data rdata;
-	struct wlr_box geom;
+	struct wlr_box geom, pg;
```
alex-courtis, line 1200: /nit naming consistency

Perhaps `pgeom` or rename `geom` to `g`.

alex-courtis:
This change is technically sound, pending fixes for chromium/intellij behaviour.

I'm not sure that this change will result in a good user experience: cutting off menus etc. will make tiled/floating applications difficult to use.

Perhaps this one should be an option in the config: it's baked too deep for a patch, and will be relevant for the majority of users.

djpohly:
Yeah, I just wanted to hack this up for some testing. I do really think clients should be kept in their own space, with perhaps a small number of well-defined exceptions. (I particularly don't like giving X11 clients free rein of my Wayland screen just because "they don't know better."  It's like the neighbors trying to excuse their dog for pooping in your flowerbed because they think fences are bad UX.)

Sevz17:
This pr does not longer applies

Issue #7, opened 2020-06-09
SUID: Drop root permission like Sway


sdsddsd1:
Hey,
since this project might be  interesting for minimal envoriments without systemd/elogind, I wanted to ask if its on your agenda to add support for setuid on the binary and drop the provilieges as soon as they are unneded like it is done in Sway.
[sway wiki](https://github.com/swaywm/sway/wiki/Running-Sway-without-systemd#direct-via-setuid)
[Sway refuses to start when it cant drop privileges](https://github.com/swaywm/sway/blob/33a984bbc5c5ec4202f24bacf6ddd584b95a9ec0/sway/main.c#L187)

djpohly:
Sounds good to me.  Send a PR to review?

sdsddsd1:
I am not able to do this. I just want to ask or encourage the conversation.

LukasZumvorde:
A pull request has been added trying to copy this feature form Sway #36

djpohly:
Can someone with this use case test the PR?

sdsddsd1:
I would test, but I cant build dwl because of missing `xwayland`.   I cannot find a way to disable it in the makefile.

alex-courtis:
xwayland is not optional; you'll need it to build and run.

djpohly:
It ought to be optional, really.  You shouldn't have to install X11 cruft in order to run dwl.

djpohly:
Fixed that. @sdsddsd1, give it another try!

sdsddsd1:
I have no success to start dwl. It cannot become 'drm master' and ends with a segmention fault. There is `chmod u+s /usr/bin/dwl`.  
Here is a strace, if that helps. https://termbin.com/195v

sdsddsd1:
@LukasZumvorde , I somehow couldnt ping you in the first post.

sdsddsd1:
Is this feature still relevant? As of linux 5.8 it is possible to become and drop drm_master for unprevileged users. This feature is available in the latest wlroots master. I can run other compositors with setgid on input as a user process.
```
chown :input /usr/bin/sway
chmod g+s    /usr/bin/sway
```

I can start also dwl like this, but firefox refuses to start because of allegedly root permissions.

djpohly:
@sdsddsd1 Do you still have an issue with this on your setup, or did the Linux fix (plus, perhaps, adding your user to the `input` group) solve it?

sdsddsd1:
[...]

PR #36, opened 2020-08-07
SUID: Drop root permission like Sway

LukasZumvorde:
This is implementing the permission dropping feature like it is done in sway.
 
I have just stumbled over this project and don't use dwl as a daily driver yet. Therefore it has not been tested in real world conditions. The feature has been tested in a fresh VM though.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -671,6 +671,25 @@ dirtomon(int dir)
 	}
 }
 
+static bool drop_permissions(void) {
+	if (getuid() != geteuid() || getgid() != getegid()) {
+		// Set the gid and uid in the correct order.
+		if (setgid(getgid()) != 0) {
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -1846,6 +1865,10 @@ main(int argc, char *argv[])
 		exit(EXIT_FAILURE);
 	}
 
+	if (!drop_permissions()) {
```
djpohly, line 1868: I would guess from the error posted over at #7 that root is needed at least until `wlr_backend_autocreate` returns.

travankor:
Is this relevant since you can use [seatd](https://sr.ht/~kennylevinsen/seatd/) now?

Sevz17:
Sway deprecated SUID (https://github.com/swaywm/sway/commit/e5728052b59fb5b476c78c9f18b812a85d7f4503)

Issue #35, opened 2020-08-04
Use cage for X11 clients


djpohly:
Using [cage](https://github.com/Hjdskes/cage) to run X11 clients could have a lot of benefits - cleaner code, more control over client isolation, and so on.

pm4rcin:
1. Is cage capable of running multiple X11 programs at the same time?
2. What about programs like e.g. Lutris/Bottles that are Wayland native but they run X11 games. Is cage able to take the game from them?
2.1 What if the game has a launcher? Since cage is able to run one application will it work?

godalming123:
If your asking for my opinion, as a DWL user that removed xwayland support from their fork of DWL I think this is a good idea but still allows to run xwayland clients if necersarry 👍, I just can't speak for the rest of the community.

pm4rcin:
@sevz17 It looks like cage is now back to being maintained by emersion himself so I wouldn't worry too much about it being abandoned. I understand that's not an easy decision to make and many things should be taken into consideration. But I'd like hear what do you think about that?

sevz17:
Ah, yeah, I haven't had time to delve into details, but the memory footprint worries me a bit.

PR #37, opened 2020-08-09
31 independents mouse focus

alex-courtis:
Independent gets focus if the mouse is over it, otherwise client as usual.

"Focus" is `wlr_seat_pointer_notify_enter` and nothing else.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -224,6 +224,7 @@ static void unmapnotify(struct wl_listener *listener, void *data);
 static void xwaylandready(struct wl_listener *listener, void *data);
 static void view(const Arg *arg);
 static Client *xytoclient(double x, double y);
+static Client *xytoindependent(double x, double y);
 static Monitor *xytomon(double x, double y);
 static void zoom(const Arg *arg);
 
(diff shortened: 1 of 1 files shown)


djpohly:
Unmanaged windows in X11 are never supposed to receive actual focus, so I think the wlr-pointer-focus-only makes sense.

However, I am starting to think that XWayland should become an external patch.  Supporting X11 stuff is really starting to make the code suck (not commenting on the quality of your implementation, just all the hoops and extra cases we have to jump through) for a use case that not everyone needs.

alex-courtis:
> Unmanaged windows in X11 are never supposed to receive actual focus, so I think the wlr-pointer-focus-only makes sense.
> 
> However, I am starting to think that XWayland should become an external patch. Supporting X11 stuff is really starting to make the code suck (not commenting on the quality of your implementation, just all the hoops and extra cases we have to jump through) for a use case that not everyone needs (see also #35).

The `#ifdef XWAYLAND` worked out better than I thought it would.

Not sure a patch will work for xwayland - it will very quickly become stale while we are developing dwl. Perhaps once dwl is stable.

PR #40, opened 2020-08-14
Fix Firefox class

guidocella:


djpohly:
Thanks - I'm actually going to comment these out as well, since they're only meant to be examples.

PR #51, opened 2020-08-27
Allow borderpx=0

guidocella:
Avoids crashing with the error dwl: render/wlr_renderer.c:102:
wlr_render_rect: Assertion `box->width > 0 && box->height > 0' failed.


djpohly:
Could we try a PR to wlroots to return instead of assert in this case?  Zero is a valid dimension imo. (In theory there are sensible things you can do with negative dimensions too, but that's less intuitive.)

guidocella:
Fixed in wlroots.

PR #53, opened 2020-09-05
Fullscreen support

Stivvo:
A client can be set fullscreen in two ways:

+ From the client itself (playing a video in fullscreen on a browser or trough a shortcut built into the client)
+ From dwl with ``mod + e`` (works on all clients)

The server tells the client to lose its fullscreen status when a new client is added to the same tag. Instead, if it is moved to a different tag where there already are other clients, it keeps it fullscreen status (no titlebar...) while still being resized to fit in the new tag's layout. Since this situation is a quite rare and can easily fixed with another fullscreen toggle, I don't feel like it is a major inconvenience.

Among all the clients I tested (various qt and gtk applications for wayland, dia and steam for xwayland) the only one who doesn't play nicely with server imposed fullscreen is firefox (wayland) and I still don't understand why. It works perfectly when using f11 though.

By the way, this is an awesome project! When the layer shell protocol and fullscreen support will be completed I'll completely move from sway to dwl as a daily driver.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,6 @@ dwl is a work in progress, and it has not yet reached its feature goals in a num
 - Statusbar support (built-in or external)
 - layer-shell
 - Damage tracking
-- Fullscreen/fixed windows (or whatever the Wayland analogues are)
 
 
 ## Acknowledgements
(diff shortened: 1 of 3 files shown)


guidocella:
This doesn't support opening windows in fullscreen. Try running
`mpv --fullscreen --force-window --idle`
or
`chromium --start-fullscreen`
I think either `createnotify` should disable the fullscreen only of clients other than the one being created or this behaviour should be left as a patch.

Stivvo:
I haven't thought about that. Still, even without ``quitfullscreen`` in ``createnotify`` applications can't start in fullscreen. So, maybe this isn't the only problem.

guidocella:
I guess fullscreen at startup is part of "Client-initiated window management, such as move, resize, and close, which can be done through the compositor" which is considered a non-goal in the README. That seems to refer to the wlr_foreign_toplevel_manager_v1 protocol used in sway's `view.c`.
Nevermind then.

djpohly:
Is this still a work in progress or ready for review?

PR #54, opened 2020-09-08
Sway scale_box (rounding)

Bonicgamer:
Scaling a wlr_box without rounding can sometimes make the borders not connected and nice.
I noticed this when setting scale in monrules to 1.2

So I've went and copied what Sway did in desktop/output.c but without having a second function and now using a random rounding macro I found on the internet so as to not use round from math.h.


djpohly:
Looks good.  Do you see a need for special-casing negative numbers or using `long` instead of `int`?  If not, we can simplify the `ROUND` macro a bit.

PR #62, opened 2020-10-17
Added basic tap-to-click for touchpad users

Java-boi:
Hi, I'm using dwl on my Laptop and added tap-to-click for a smoother experience on such devices.


djpohly:
Thanks!

PR #65, opened 2020-11-22
Add install target to Makefile

Keating950:
Adds a method for easy installation. `$PREFIX` is set in `config.mk`.


djpohly:
Cool, thanks!

Issue #68, opened 2020-12-25
Monitor geometry not correct for X11 backend


djpohly:
Commit c9f92bd caused the problem. The correct way to implement may be to handle another event (sigh...) from output or output_layout.

djpohly:
Fixed by #71.

PR #71, opened 2020-12-28
call updatemons at the appropriate times

djpohly:
Addresses part of the problem of #68, but is still wonky when there are multiple outputs.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -290,7 +290,7 @@ static void toggleview(const Arg *arg);
 static void unmaplayersurface(LayerSurface *layersurface);
 static void unmaplayersurfacenotify(struct wl_listener *listener, void *data);
 static void unmapnotify(struct wl_listener *listener, void *data);
-static void updatemons();
+static void updatemons(struct wl_listener *listener, void *data);
 static void view(const Arg *arg);
 static void virtualkeyboard(struct wl_listener *listener, void *data);
 static Client *xytoclient(double x, double y);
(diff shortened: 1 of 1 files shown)


djpohly:
@Stivvo What's the code in [lines 853-879](https://github.com/djpohly/dwl/blob/67c2b74ffe45c009fc6eb735671452500785a652/dwl.c#L853-L879) meant to do?  The commit says monitor order "actually matters," but it doesn't elaborate beyond that.  It looks like the code duplicates some of add_auto, and if I replace it with the original add_auto call, it fixes the output position/size bugs I'm seeing with WLR_X11_OUTPUTS=2 on the x11 backend (shown below).

current code (two dwl outputs are the left half and the top-right quarter):
2020-12-27-224428_3200x1800_scrot

same setup reverted to add_auto:
2020-12-27-224526_3200x1800_scrot

guidocella:
https://github.com/djpohly/dwl/commit/5668c616161d451e6f20be29b31bbf03f0f398a5 makes monitors be added in the order they are defined in config.h. Stivvo released that as a patch, but I assumed it was desirable in mainline because the readme mentioned "Configurable multi-monitor layout support, including position and rotation" as a goal.

Stivvo:
The description on the wiki and in [config.def.h](https://github.com/Stivvo/dwl/blob/cd4d1a4352a4bd44805d664729c3910a20c59a92/config.def.h#L35-L37) give a better explanation of what the patch actually does.
Basically in those lines dwl is checking in which position the new monitor was configured, then is inserted to the right of the monitor that is currently nearest to that position (from the left). Then, all the x coordinates of the monitors and the floating windows on the right of the new monitor are shifted to the right by the width of the new monitor.
From my testing, this patch works starting dwl from tty or from a display manager and I cannot replicate this bug.
Yes, it was meant to be added to mainline dwl eventually but with the outputManagement patch everything that this patch does and more can be replicated using wlr-randr, the advantages of outputOrder being the simpler configuration, efficiency and persistence of the configuration after plugging or unplugging monitors at runtime.

djpohly:
Ahhh. It seems like we could accomplish something similar by adding x/y to monitor rules and letting wlroots do the heavy lifting, rather than juggling the monitors around ourselves when it's already implemented for us.

Stivvo:
I did it that way because it doesn't involve any additional change in config.h, but It is also more limited since monitors can't be arranged vertically for example while also being more complex. I've changed that in #71.

PR #72, opened 2020-12-30
Define monitor's x,y at compile time

Stivvo:
Replaces the outputOrder patch.

This avoids recalculating positions and allows to arrange monitors in
any order, not just from left to right.

The order in which monitors are defined in config.h still matters but
it's just the order in the list, not the actual position.

See #71

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -28,12 +28,14 @@ static const Layout layouts[] = {
  * The order in which monitors are defined determines their position.
  * Non-configured monitors are always added to the left. */
 static const MonitorRule monrules[] = {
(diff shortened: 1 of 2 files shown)


djpohly:
Which functionality still depends on the monitors having a specific order?

Stivvo:
It's the order in which focusmon and tagmon cycle trough the monitors.

djpohly:
Got it - I saw that it was used there but wasn't sure if there was something else I missed.  I've got an idea that might simplify things even more in that case.  Thanks for the patch?

PR #76, opened 2021-01-04
remove EGL parameter from backend_autocreate

djpohly:
Tracking with breaking change in swaywm/wlroots#2593.


emersion:
Side note, this is actually https://github.com/swaywm/wlroots/pull/2561, not https://github.com/swaywm/wlroots/pull/2593.

PR #77, opened 2021-01-06
Determine monitor order spatially

djpohly:
Cycling through monitors in left-to-right/right-to-left order is reasonable behavior, and wlroots already provides this functionality, so let's not reimplement anything we don't have to.

This should make it easier to bring back autoconfigured outputs.  It has the added benefit of putting the output-order logic in a single, well-defined location (easily hackable/patchable!) for anyone who prefers a different default.


Stivvo:
Well, it makes sense and it is a lot cleaner.

PR #80, opened 2021-01-11
Fix Initialization of NetWM Atoms

sam-barr:
The initialization of the utility window type and the toolbar window type were swapped.

Additionally, variables xcursor and xcursor_mgr are only used
when xwayland is defined, so I make the variables declaration
contingent on whether xwayland is being used


djpohly:
Looks good.  Thanks!

PR #81, opened 2021-01-12
Setting monitor modes in config

meutraa:
### Description

Allow setting custom modes in the config file.

If set to 0, 0, 0, will use monitor's preferred mode.
```
{ "eDP-1",    0.5,  1,      2,    &layouts[0], WL_OUTPUT_TRANSFORM_NORMAL, 0, 0, 0, 0, 0 },
```

Otherwise, if all three values are non-zero, 
1. If monitor has no modes, set using custom mode function.
1. Loop through monitor modes
    1. If exact match, use that mode.
    1. If no exact matches, but resolution matches, use last mode in matching list (I suspect the highest refresh rate).
1. Fallback to custom mode.

I suspect the logic could be shortened a little, since if step 1 is skipped, there should be an empty loop, and the fallback will trigger. But my C is rusty.

### Motivation

I have monitors with preferred modes 1920x1080@144 and 3840x2160@30, and I personally prefer 1920x1080@240 and 1920x1080@60 respectively.

### Disclaimer

I have almost no wayland knowledge, and I have not touched C in a long time.

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -28,9 +28,9 @@ static const Layout layouts[] = {
  * The order in which monitors are defined determines their position.
  * Non-configured monitors are always added to the left. */
 static const MonitorRule monrules[] = {
-	/* name       mfact nmaster scale layout       rotate/reflect x y */
+	/* name       mfact nmaster scale layout       rotate/reflect              x  y  w     h     rate */
 	/* example of a HiDPI laptop monitor:
-	{ "eDP-1",    0.5,  1,      2,    &layouts[0], WL_OUTPUT_TRANSFORM_NORMAL, 0, 0 },
+	{ "eDP-1",    0.5,  1,      2,    &layouts[0], WL_OUTPUT_TRANSFORM_NORMAL, 0, 0, 1920, 1080, 60000 },
 	*/
 	/* the order in which monitors are defined here affects the order in which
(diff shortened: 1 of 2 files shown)


djpohly:
Hm, this is a lot of lines added to implement something that seems like an uncommon case - I'd assume a monitor's preferred mode is usually a sane default.  Definitely worth being a patch that's out there for those who need it, but I'm not sure it goes in mainline.

meutraa:
I think that is a naive assumption. Especially given rotation, dpi, and position is already implemented. 

If you want to reduce the lines of code, I could remove fallbacking and custom modes, and inline the function (like the rest of the monitor rule setup). But ~20 SLOC seems low enough honestly. The user might get frustrated when their 240hz monitor is actually 239.760.

djpohly:
I'm thinking about pushing some of the other monitor config out to patches as well, so please don't feel like this is singled out.  Supporting the output-management protocol allows tools like `wlr-randr` to be used for this purpose.  (I would actually love to get back to using wlroots autoconfig by default rather than implementing our own, but some other changes would need to happen first.)

The preferred mode *should* be a sane default - if not, there's an issue that might be worth reporting: either the kernel DRM driver is tagging the wrong mode as `DRM_MODE_TYPE_PREFERRED` or the hardware is providing incorrect EDID information.

PR #83, opened 2021-01-13
Add a nix-shell file to create a build environment

meutraa:
### Description

This adds a `default.nix` file that is read by the program nix-shell. It creates a shell environment with the required dependencies to build dwl. I also added a little instruction to the README.md file.

### Motivation

I had to fork nixpkgs to install a more recent version of wlroots so that I could build dwl. Since configuring dwl requires building, having nix-shell set up the build environment with library dependencies saves me a lot of time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,12 @@ dwl has only two dependencies: wlroots-git and wayland-protocols. Simply install
 
 To enable XWayland, you should also install xorg-xwayland and uncomment its flag in `config.mk`.
 
+### Nix
+
+If `config.mk` has the xwayland flag, set `enableXWayland` to `true` in `default.nix`.
+
+Run `nix-shell --pure` to enter the shell, and then `make` to build dwl.
+
 ## Configuration
 
 All configuration is done by editing `config.h` and recompiling, in the same manner as dwm. There is no way to separately restart the window manager in Wayland without restarting the entire display server, so any changes will take effect the next time dwl is executed.
--- a/default.nix
+++ b/default.nix
@@ -1,8 +1,8 @@
 { pkgs ? import <nixpkgs> { } }:
 let
   enableXWayland = false;
-  version = "4f80fab337f10b4ad2043b834606540895b8df36";
-  sha256 = "0zsfglyfmzsxf6vkdv999z49v66pwcyffb0pn824ciafwna2408r";
+  version = "dc61f471da1a1c9264167635c286b6dcb37b3d6f";
+  sha256 = "0k31chpc4facn7n7kmk0s5wp7vj7mpapwk4as6pjhi1rq37g34lf";
 
   wlroots-git = pkgs.wlroots.overrideAttrs (old: {
     version = version;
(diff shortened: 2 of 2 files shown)


meutraa:
Just as a test, I built it on a headless raspberry pi 4 to make sure it works on other systems.

theotherjimmy:
> I had to fork nixpkgs to install a more recent version of wlroots

I'm seeing a modified version of the wlroots derivation. Is that what you meant?

meutraa:
> > I had to fork nixpkgs to install a more recent version of wlroots
> 
> I'm seeing an overridden version of the wlroots derivation. Is that what you meant?

I mean, no, that was just my motivation for this change - to save me the effort of setting up the build environment each time.

theotherjimmy:
Right, I somehow missed the large word above the text 🤦

djpohly:
I'm not convinced it's a good idea to set a precedent of using this repo to maintain the packaging and environment configuration for all the various distros and tools out there.  Does NixOS/nix-shell have something like Arch's AUR where these user-contributed configurations can go?

meutraa:
> I'm not convinced it's a good idea to set a precedent of using this repo to maintain the packaging and environment configuration for all the various distros and tools out there. Does NixOS/nix-shell have something like Arch's AUR where these user-contributed configurations can go?

This does not package, it sets up a required build environment. I feel like you could make the same argument for the Makefile. It is not source code, it runs the required commands to build the application - minus all the commands required to create the environment to build.

djpohly:
I appreciate that nix is a neat tool, but dwm doesn't use anything fancy to set up a build environment; the user is expected to be able to install the required dependencies, edit the config, apply patches, and so forth. I'm trying to keep this project something that even stauncher suckless types would consider hacking on, and it would be an immediate turn-off to tell them they need to install a package manager they don't have so that they can use one of its auxiliary commands to install dependencies that they could have installed themselves (in less time than it took to find, download, and build said package manager).

I understand it's frustrating that your distro doesn't provide an easy way to build wlroots-git, but this isn't the place to fix that.  You're 100% welcome to add this to the patches page on the wiki so that others who need it can find it... I'm just not too keen on making it part of the official repo or build process.

meutraa:
> and it would be an immediate turn-off to tell them they need to install a package manager they don't have so that they can use one of its auxiliary commands to install dependencies that they could have installed themselves (in less time than it took to find, download, and build said package manager).

This does not make nix a build dependency. It just makes the environment setup automatic for those who already use nix.

djpohly:
If you can convince the dwm maintainers that they need a nix-shell file in the suckless.org repo, then please definitely re-open this and I'll give it some serious consideration.  For now, I'm going to hold the line on not maintaining the configuration for everyone's various toolchains... thanks for understanding.

Fuseteam:
btw @meutraa wouldn't it also be fixed if the wlroots in nixpkgs itself is updated? maybe a PR from your fork to upstream might help?

PR #84, opened 2021-01-13
Add config.h to .gitignore

meutraa:
### Motivation

`config.h` is effectively build state, this stops anyone accidentally committing it.


djpohly:
Makes perfect sense.

PR #87, opened 2021-01-16
WIP: Basic damage tracking

### LINKED ISSUE #85: Reduce monitor rendering
DWL appears to be redrawing the framebuffer on every refresh. On my test NixOS VM running under libvirt with a default configuration, each of the 8 cores has ~65% CPU usage. With this patch applied, the CPU usage goes to near 0 when there are no clients.

```
diff --git a/dwl.c b/dwl.c
index c09f598..bb1315a 100644
--- a/dwl.c
+++ b/dwl.c
@@ -1759,7 +1759,7 @@ void
 rendermon(struct wl_listener *listener, void *data)
 {
 	Client *c;
-	int render = 1;
+	int render = 0;

 	/* This function is called every time an output is ready to display a frame,
 	 * generally at the output's refresh rate (e.g. 60Hz). */
@@ -1770,6 +1770,7 @@ rendermon(struct wl_listener *listener, void *data)

 	/* Do not render if any XDG clients have an outstanding resize. */
 	wl_list_for_each(c, &stack, slink) {
+		render = 1;
 		if (c->resize) {
 			wlr_surface_send_frame_done(client_surface(c), &now);
 			render = 0;
```

This is not a pull request, just a test case to demonstrate the issue. This very quick and dirty patch disables drawing unless there is a client.

MilesBreslin:
Resolves #85. This is only basic support for damage tracking and is implemented so that the entire monitor will be damaged for each client frame commit.

This has only been tested in a VM and needs hardware testing.


MilesBreslin:
I did some further testing and, it does not seem to work for anything besides toplevel surfaces. Since the `createnotify` only creates a `Client` for toplevel surfaces.

This means Xwayland and popups are not supported. I could use some help with these since I'm new to wayland programming. If anyone can point me in the right direction, I'd love to get this feature completed.

dwl.c
```diff
@@ -770,6 +773,10 @@ commitlayersurfacenotify(struct wl_listener *listener, void *data)
 			&layersurface->link);
 		layersurface->layer = wlr_layer_surface->current.layer;
 	}
+
+	// Damage the whole screen
+	if (selmon)
+		wlr_output_damage_add_whole(selmon->damage);
```
djpohly, line 779: I think you mean `m` instead of `selmon`, right?  I originally intended that neither `selmon` nor `c->mon` nor `output->data` would ever be null - it's possible that the output-management changes affected that invariant, but if not, there's no need to do this conditionally.

dwl.c
```diff
@@ -780,6 +787,10 @@ commitnotify(struct wl_listener *listener, void *data)
 	/* mark a pending resize as completed */
 	if (c->resize && c->resize <= c->surface.xdg->configure_serial)
 		c->resize = 0;
+
+	// Damage the whole screen
+	if (c->mon)
+		wlr_output_damage_add_whole(c->mon->damage);
```
djpohly, line 793: Can `c->mon` ever be null if we get here?

dwl.c
```diff
@@ -1018,6 +1030,11 @@ destroynotify(struct wl_listener *listener, void *data)
 {
 	/* Called when the surface is destroyed and should never be shown again. */
 	Client *c = wl_container_of(listener, c, destroy);
+
+	// Damage the whole screen
+	if (c->mon)
+		wlr_output_damage_add_whole(c->mon->damage);
+
```
djpohly, line 1037: (see above re: `c->mon`)

dwl.c
```diff
@@ -1360,6 +1377,10 @@ mapnotify(struct wl_listener *listener, void *data)
 	/* Set initial monitor, tags, floating status, and focus */
 	applyrules(c);
 
+
```
djpohly, line 1380: Extra blank line.

dwl.c
```diff
@@ -1769,7 +1790,14 @@ rendermon(struct wl_listener *listener, void *data)
 	if (!wlr_output_attach_render(m->wlr_output, NULL))
 		return;
 
-	if (render) {
+	bool needs_frame;
```
djpohly, line 1793: Suckless style: declarations before statements, `int` not `bool`, one-line blocks without braces where not confusing.

djpohly (commented):
Thanks for the start on this - appreciate trying the minimalist/good-enough approach first to see if it will work.

As for Xwayland and popups, we just need to know when those surfaces are committed and what outputs they are on, right?

dwl.c
djpohly: Another note here: we can probably remove `wlr_output_attach_render`, since that's the first thing that `wlr_output_damage_attach_render` calls.

djpohly: (Oh, I see about the `bool`. Ugh, thanks "unopinionated" wlroots.)

djpohly:
Hmmm, now that I think about this, we would need to damage any monitor which intersects with the surface's box.  What if we just consider all outputs damaged after a surface commit?  It should still cut down significantly on spurious renders, and we might even be able to keep track of it separately from wlroots and Pixman.

dwl.c
MilesBreslin: This solved an out of bounds panic I was encountering. If we can guarantee that `commit` is never called before a surface is mapped, then this wouldn't be necessary, right?

dwl.c
MilesBreslin: I'm unfamiliar with the suckless coding style and actually any C coding standards for that matter. I'm happy to change this, but wlroots does use a bool here. Leaving for as is now.

MilesBreslin: Removed the `wlr_output_attach_render` a new revision.

dwl.c
MilesBreslin: Ah, completely missed that. Resolved in new revision.

MilesBreslin:
Latest commit does not use any wlroots damage tracking features. Just an `isdamaged` field per monitor. This implementation has a quite obvious uglyness that probably could be fixed by either variable naming or better buffer management. Each time a surface is damaged, we must redraw twice since there are 2 buffers being swapped. I can't roll this one back because it seems to force the screen black when I do `wlr_output_rollback`. Without rendering the frame twice, the screen will flicker.

dwl.c
djpohly: Ah, you're right, the surface can be committed before attaching a buffer (which would cause it to be mapped on the following commit).  This makes sense then... and now, having thought it through, seems like adding `&& VISIBLEON(c, c->mon)` would be even better.

djpohly:
You make a good point - it's probably best to let wlroots handle the intricacies of double-buffering.  Turns out your previous implementation allows us to make the rendering logic a touch more straightforward.  I've added a few commits to move in that direction.

Only damaging a single output isn't quite enough - say you're mirroring screens, for example.  I think we'll have to just damage them all when there is a change.

Stivvo:
Firefox and discord stutter (especially playing videos), like they were rendering only once a second. On firefox, this happens only when interacting with the webpages, not other gui elements like the tabs. Honestly I have no idea why this is the case. 
All other applications I tested work fine and the reduction in CPU load is very impressive.

MilesBreslin:
My theory on firefox is that it uses more than one surface which we are not catching the commits for. This only supports top-level surfaces at the moment.

Stivvo:
The fix for the issue I mentioned a few days ago is now ready as a pull request to  the MilesBreslin's branch.

MilesBreslin:
Seems to work great! Popups still don't appear to work, such as the hamburger menu in firefox. XWayland also does not work. Both do not trigger rendering.

drdonahue:
I noticed an issue, when I send a client to a different monitor, the monitor that the client was sent from from stops rendering until a new client is spawned on it.

Stivvo:
It seems like when a client is moved to another monitor, himself and the clients on the destination monitor are resized. Their ``resize``  is sometimes set to != 0 if the system can't keep up (chances increase  with more clients, in fact this issue rarely appears moving to an empty monitor). This makes  ``rendermon()`` skip a few frames, untill there are no pending resizes. However, by then ``wlr_output_damage_add_whole()`` had already been called, so that monitor won't be rendered untill something else happens on it. Deliting [these lines](https://github.com/djpohly/dwl/blob/3a72cd924c45d4bc01f4d8477c14fdf2854f2dd0/dwl.c#L1751:L1758) and everything regarding ``Client.resize`` fixes the problem.
By the way, I don't get how ``Client.resize`` was useful in the first place, whithout it dwl feels smoother and there is less overhead.

Stivvo:
Also https://github.com/MilesBreslin/dwl/pull/3 simplifies things a bit.

MilesBreslin:
I'm glad to see this PR continue moving. I have merged @Stivvo 's changes. I have looked into this some again, but unfortunately, I can't seem to figure out the popup issues. It seems to me that that is the last remaining problem with this feature. I appreciate the help, and I think I will need some more getting this last bug figured out.

As I understand it, we are not catching the `commit`s from the popup windows. We either need to add a listener or use an existing listener (if it is under one) to catch the `commit` signals from the popup. Once caught, we just need to damage the whole display.

As for the `Client.resize` I thought that was there because on a resize, a client will be resized on the layout, but the buffer may still be the incorrect size. This would force us to wait until the client has resized to the appropriate resolution for its place on the display. Although as discovered previously in the PR, skipping rendering this way should cause some bugs anyway.

Stivvo:
Which bug relative to popups do you actually see? I haven't noticed anything wrong so far. Same thing with removing ``Client.resize`` (although I got the idea behind it), but I'm probabilly missing something.

MilesBreslin:
I've been having issues with the Firefox hamburger menu not refreshing. Steps to reproduce, open Firefox, click the menu, and hover over different menu options. It should indicate which menu item you are hovering over, but does not in my testing. I have been testing as a subcompositor in Sway.

I have not changed `Client.resize` at all, but in one of the earlier revisions, I experienced issues with reimplementing damage tracking as `Client.is_damaged`, which worked in a similar way. After a period of time without a refresh, the screen would blank on my VM (QEMU defaults, with dwl as the only compositor). If I made the screen refresh without redrawing, the screen would alternate between 2 buffers, one which was out of date.

mortezadadgar:
So I merged these patches into my personal build and CPU usage went from 7% down to 1% which is great, the only issue was that firefox got utterly slow and difficult to use

notthewave:
So I tried out this pr and firefox (89) works fine. But the foot terminal causes some problems. 
- if I spawn multiple instances of foot it flashes 
- when I close foot dwl crashes

Sevz17:
with MilesBreslin#4 the crash with foot is solved

notthewave:
MPV still lags when the video is playing.

notthewave:
On resize, I might add...

ARDiDo:
MilesBreslin#6 should keep track of subsurfaces in XWayland clients.

PR #88, opened 2021-01-25
Add reference to IRC/Matrix room.

DanielMowitz:
Since we also have a matrix/IRC community, it shouldn't go unnoticed.


DanielMowitz:
The proposed change of the README has now been implemented in a different way.

PR #91, opened 2021-02-07
Interface to display tag information on status bar

drdonahue:
I would like to add in a way to replicate the tag, layout, and client title information that DWM provides in its bar. I've implemented a solution using a file in /tmp as a proof of concept; I would love feedback on how to handle the interface. The file this creates can be parsed by a simple script and added as a waybar module.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -2200,6 +2200,7 @@ tag(const Arg *arg)
 		focusclient(focustop(selmon), 1);
 		arrange(selmon);
 	}
+    statusbar();
 }
 
 void
@@ -2269,6 +2270,7 @@ toggletag(const Arg *arg)
 		focusclient(focustop(selmon), 1);
 		arrange(selmon);
 	}
+    statusbar();
 }
 
 void
@@ -2281,6 +2283,7 @@ toggleview(const Arg *arg)
 		focusclient(focustop(selmon), 1);
 		arrange(selmon);
 	}
+    statusbar();
 }
 
 void
@@ -2361,6 +2364,7 @@ view(const Arg *arg)
 		selmon->tagset[selmon->seltags] = arg->ui & TAGMASK;
 	focusclient(focustop(selmon), 1);
 	arrange(selmon);
+    statusbar();
 }
 
 void
--- a/dwl.c
+++ b/dwl.c
@@ -280,6 +280,7 @@ static void setmon(Client *c, Monitor *m, unsigned int newtags);
 static void setup(void);
 static void sigchld(int unused);
 static void spawn(const Arg *arg);
+static void statusbar(void);
 static void tag(const Arg *arg);
 static void tagmon(const Arg *arg);
 static void tile(Monitor *m);
@@ -1116,6 +1117,7 @@ focusclient(Client *c, int lift)
 		wl_list_insert(&fstack, &c->flink);
 		selmon = c->mon;
 	}
+    statusbar();
 
 	/* Deactivate old client if focus is changing */
 	if (old && (!c || client_surface(c) != old)) {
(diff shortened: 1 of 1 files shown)


djpohly:
Cool - this is about the simplest way to approach it.  Two thoughts:

1. How about stdout?  I don't see any other use dwl would have for it.
2. What would you think about making the format more extensible (while still being line-oriented and easy to parse), so that if someone wants to add a patch that provides additional info, it continues to work with existing setups?  Something along the lines of:
    ```
    selmon eDP-1
    title eDP-1 dwl.c (~/src/dwl) - VIM
    tags eDP-1 11 2
    layout eDP-1 []=
    title HDMI-1 GitHub - Firefox
    tags HDMI-1 3 1
    layout HDMI-1 TTT
    ```

and then you have a variety of options like:
```
dwl > /tmp/dwltags
dwl | ./format-for-mybar | mybar
```

drdonahue:
stdout should work fine, though one potential issue is that processes spawned with -s also print out to the TTY. I need to check whether they are printing to stdout or just stderr, but that could be a conflict. Assuming that's not an issue, stdout is the most flexible option.

I like the format you're suggesting. I think it would be best to output the display name first instead of the field name, because if you have a process per output that's what they're going to want to filter on first, but that is a lot better than my mock-up for extensibility.

drdonahue:
I take that back, spawned processes are not outputting to the TTY. I must have been seeing something else. stdout is clear, so I'll switch over to that.

djpohly:
I still think you're right to be cautious about the open file descriptor, but it doesn't have to keep us from using stdout. The "right" fix is probably to dup some things to /dev/null, but I bet we could get away with a judiciously-placed

    dup2(STDERR_FILENO, STDOUT_FILENO);

drdonahue:
OK, that has it working with the new output format via stdout.

drdonahue:
OK, waybar *does* send stdout and stderr through when ran with -s, not sure why I couldn't reproduce it the other day. I'm going to look into a redirection like you suggested, or that's going to break things.

drdonahue:
There are the judiciously placed dup2's. It wouldn't be hard to switch it over to a /dev/null sink if that's the better way to do it, but either way leaves stdout free.

djpohly (changes requested):
Great!  Two more nits:
1. `fprintf(stdout, ...)` -> `printf(...)`
2. How about calling the function `printstatus` or `printstate` or such, since it doesn't have to be used for a bar?

drdonahue:
OK, that should resolve the nits.

djpohly (approved):
Great work. Thanks!

ahmadraniri1994:
Interesting, how do we use it, can we apply it to waybar?

michaelskyba:
> Interesting, how do we use it, can we apply it to waybar?
> Update: found a way, thanks

How'd you do it? I can't find any documentation for this

ahmadraniri1994:
@michaelskyba though I am not currently use dwl right now, I will suggest to use yambar and dwl script. Here, my last dwl + yambar snapshot, I can see the current workspace and dwl mode : 
yambar-dwl

drdonahue:
> > Interesting, how do we use it, can we apply it to waybar?
> > Update: found a way, thanks
> 
> How'd you do it? I can't find any documentation for this

I don't use waybar anymore, but what you do is add a custom module in your waybar config such as:
```
    "custom/dwlbar": {
        "format": "{}",
        "return-type": "json",
        "max-length": 200,
        "exec": "$HOME/.config/waybar/dwlbar 'HDMI-A-1'",
        "escape": true
    }
```
And handle the tag info in the 'exec' script that you give it.

PR #112, opened 2021-04-19
Add uninstall target in makefile like dwm

JaGoLi:
This is just a really simple commit that allows users to run the "make uninstall" command to remove dwl, like it is possible for dwm.


djpohly:
Thanks!  (I feel obligated to note here that one should never use `make install` and `make uninstall` to manage system-wide installation, but they can be useful for non-admin users who wish to install to `$HOME/.local` or such.)

PR #121, opened 2021-05-23
implement urgency hint

djpohly:
Implements urgency hints using the xdg-activation-v1 protocol.  This should be ready to merge once swaywm/wlroots#2718 lands.


djpohly:
Merged to wlroots-next branch and will hit main with the next wlroots release.

PR #122, opened 2021-05-23
before set tiled verify if client is xdg-shell, then set tile

Sevz17:
The pull request #120 causes that when you want to start a xwayland client, dwl crashes, this change solves it


djpohly:
Thanks for the fix!

PR #127, opened 2021-06-06
fix crash when a client opens in full screen mode

Sevz17:
try to open a client in fullscreen mode (example: foot -F) causes that
dwl crash, this prevent crash but this isn't a solution, since client will
be open in normal mode, we must find a way to set c->mon in createnotify
function

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1934,6 +1934,9 @@ setmon(Client *c, Monitor *m, unsigned int newtags)
 {
 	Monitor *oldmon = c->mon;
 
+	if (m)
+		c->tags = newtags ? newtags : m->tagset[m->seltags]; /* assign tags of target monitor */
+
 	if (oldmon == m)
 		return;
 	c->mon = m;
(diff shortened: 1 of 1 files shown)


djpohly:
Hm... the intent was for `c->mon != NULL` to be invariant. Can we simply rearrange a couple of lines somewhere further up the call stack rather than tossing in an extra check?

dwl.c
```diff
@@ -888,6 +888,9 @@ createnotify(struct wl_listener *listener, void *data)
 	c->surface.xdg = xdg_surface;
 	c->bw = borderpx;
 
+	/* set monitor, then it will be overwritten by applyrules */
+	setmon(c,selmon, 0);
```
Sevz17, line 892: I think calling `applyrules` here is better than `setmon`, but it can't get alacritty's appid so I have to call setmon and modify it to assign tags even if it is the same monitor

tleydxdy:
currently this patch crashes at setmon in createnotifyx11 because the client doesn't have a surface. i.e. c->surface.xwayland->surface is NULL

Sevz17:
Closed in favor of #171

PR #129, opened 2021-06-21
build: properly link against rt

simeonschaub:
This is especially relevant for cross-compilation, since dwl uses
`clock_gettime`.


simeonschaub:
Bump. Does this look reasonable?

PR #130, opened 2021-07-01
Fix excessive printstatus() calls when dmenu is up

drdonahue:
This fixes an issue where printstatus was being called on every frame when dmenu and similar clients were open, which led to excessive resource use and/or statusbars crashing from the excessive input.


djpohly:
Thanks!

PR #132, opened 2021-07-03
implement input-inhibitor protocol

### LINKED ISSUE #123: Implement input-inhibitor protocol (support screen lockers)
We now have [swaylock](https://github.com/swaywm/swaylock) and [waylock](https://github.com/ifreund/waylock) for screen locking in wayland. I thought it would be worth opening an issue to track support for these. At the moment, I'm hopping to a new tty and using `vlock`, but I don't really know what the security implications are.

Note that I never used dwm so I don't know if this should be part of dwl at all, just putting it out there.

It seems that the input inhibitor protocol needs to be implemented? Or is that a job for wlroots? I'm still figuring out the wayland stack...

```
2021-05-26 16:42:39 - [swaylock-1.5/main.c:1186] Compositor does not support the input inhibitor protocol, refusing to run insecurely
```

Sevz17:
Close #123

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@ dwl is not meant to provide every feature under the sun. Instead, like dwm, it s
 - Idle-inhibit protocol which lets applications such as mpv disable idle monitoring
 - Provide information to external status bars via stdout/stdin
 - Urgency hints via xdg-activate protocol
+- Support screen lockers via input-inhibitor protocol
 - Various Wayland protocols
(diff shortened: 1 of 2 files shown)


xi:
I tested this locally with swaylock and it works like a charm!

I was wondering though if it would be better to move the check to `keybinding()` so that it is easier to implement key bindings that work regardless of the input lock (like `bindsym --locked` in sway).

Magnitudexx:
Is this going to get commited or does something nedd to be changed?

PR #133, opened 2021-07-16
implement idle inhibitor protocol

Sevz17:


xi:
For me this crashed when using with the idle inhibitor module in waybar.

I am also not convinced that inhibitors should only be active when they are visible. I know that sway has different kinds of inhibitors (open, visible, fullscreen, focused) but I believe for a simple compositor like dwn the mere existence of an inhibitor should be enough.

So the better place to re-enable idle would be in `idle_inhibitor->events.destroy`.

Sevz17:
> I am also not convinced that inhibitors should only be active when they are visible. I know that sway has different kinds of inhibitors (open, visible, fullscreen, focused) but I believe for a simple compositor like dwn the mere existence of an inhibitor should be enough.

certainly it would be simpler to activate the inhibitors just for existing, that should also solve the crash with waybar.
I'll make inhibitors work just by existing, unless @djpohly says otherwise

Sevz17:
> So the better place to re-enable idle would be in `idle_inhibitor->events.destroy`.

I took a shortcut here, I considered that when an inhibitor is destroyed:
1. closing a window will call `focusclient ()`
2. after closing a browser tab and then clicking something will call `focusclient ()`

PR #136, opened 2021-08-03
implement the presentation time protocol

guidocella:
This lets applications, such as mpv with --video-sync=display-resample,
know accurately when frames are displayed and ensure smooth video
playback.


djpohly:
Awesome, thanks!!

PR #143, opened 2021-08-20
keep floating clients on top

### LINKED ISSUE #75: Avoid overlaping of tiled and floating windows (like dwm)
Screenshot from 2021-01-04 10-42-05

I'm guessing this is now doable as layershell is implemented.

guidocella:
Fixes #75.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -480,6 +480,11 @@ applyrules(Client *c)
 					mon = m;
 		}
 	}
+
+	// Insert floating clients at the beginning of stack, and tiled ones at the
+	// end.
+	wl_list_insert(c->isfloating ? &stack : stack.prev, &c->slink);
+
 	setmon(c, mon, newtags);
 }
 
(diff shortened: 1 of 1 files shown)


guidocella:
This is the other possible implementation: https://github.com/guidocella/dwl/commit/50c4f500b44bdf377ff8ff300fd2c862613bfc5d

djpohly:
Thanks for working this out! Hard-coding a check for monocle into focusclient and then calling focusclient from monocle seems kind of hack-ish... could that be done more cleanly?

guidocella:
> Hard-coding a check for monocle into focusclient and then calling focusclient from monocle seems kind of hack-ish... could that be done more cleanly?

Note that the check in `focusclient()` isn't because it's called from `monocle()`, but for when you press `MODKEY+{j,k}` while in monocle.

Calling `focusclient()` in `monocle()` is just to avoid a couple of lines for checking if no client is present. It's equivalent to:

```c
Client *sel = selclient();
if (sel) {
    wl_list_remove(&c->slink);
    wl_list_insert(&stack, &c->slink);
}
```

I will switch to this if you find it more explicit.

notthewave:
Just tested the patch and found 1 problem:
- in floating mode, clients are now drawn below the currently active window instead of above.

guidocella:
Fixed.

notthewave:
Not sure if that's much of a problem, but when cycling through windows, it goes through them in the order in which they were spawned and not in the order in which they are arranged. Not sure how DWM does this.

**However, this is probably out of scope of this pr as this also happens in floating mode.**

PR #144, opened 2021-08-20
implement drag and drop

### LINKED ISSUE #134: Cannot drag selected objects like I can in Sway or X
I'm not sure if this is a bug or a feature request. I can't drag objects, such as links, text, or images with my mouse. Programs that expect this functionality include Ungoogled Chromium, LibreWolf, and various electron-based apps.

I can select these items just fine, but I can't move them around.

guidocella:
For brevity, only a single drag icon at a time is supported.

Some checks are removed from pointerfocus because
wlr_seat_pointer_notify_motion() needs to be called even if a surface
isn't focused for dragging to work, and because they're redundant since
they are already performed by wlroots, and in fact they have been
removed from tinywl too in the wlroots commit f2f3df9.

Fixes #134.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -227,6 +227,7 @@ static void cursorframe(struct wl_listener *listener, void *data);
 static void destroylayersurfacenotify(struct wl_listener *listener, void *data);
 static void destroynotify(struct wl_listener *listener, void *data);
 static Monitor *dirtomon(enum wlr_direction dir);
+static void dragicondestroy(struct wl_listener *listener, void *data);
 static void focusclient(Client *c, int lift);
 static void focusmon(const Arg *arg);
 static void focusstack(const Arg *arg);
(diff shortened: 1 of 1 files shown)


Sevz17:
Hey @guidocella, I want to merge this, but needs update

guidocella:
Rebased, and this is now just one function thanks to scenegraph.

Sevz17:
drag works but no icon is rendered, you need to attach it to the scene graph in wlr_start_drag

guidocella:
Hm, I got confused by Firefox rendering its drag icon regardless. I will try to figure out how to attach it later.

Sevz17:
Check [this](https://github.com/Hjdskes/cage/blob/master/seat.c#L652-L678) and [this](https://github.com/Sevz17/dwl/blob/custom/dwl.c#L2353-L2363).

guidocella:
Updated. Was there a reason for which you reposition the drag icon on each monitor refresh instead of when the cursor moves and add `icon->surface->sx` and `icon->surface->sy` since sway and cage don't seem to do that?

Sevz17:
> Updated. Was there a reason for which you reposition the drag icon on each monitor refresh instead of when the cursor moves and add `icon->surface->sx` and `icon->surface->sy` since sway and cage don't seem to do that?

No, I just didn't remember there was motionnotify(), and yes, to fix swaywm/sway#6704

Sevz17:
Merged in 330792b1fc37db86743d7c186577776e05295e00

PR #146, opened 2021-08-22
constraint popups to monitor

xi:
This makes context menues open to the top when right clicking near the bottom edge of the screen.

I tested this on a branch based on v0.2 because I do not have a newer version of wlroots available.


Frewacom:
This seems to work fine in v0.2.1 as well, however, it does not work as expected if a monitor has non-zero x and/or y coordinates. I did some quick testing and it seems like setting `x` and `y` of the constraint box to `0` should fix it.

xi:
I do not have multiple monitors available for testing so someone else needs to take over.

PR #148, opened 2021-09-04
fix crash when borders align perfectly with screen borders

xi:


Sevz17:
I think the correct approach is to make sure that `scalebox()` never sets width and/or height < 0

PR #150, opened 2021-09-24
client_pending has been renamed as pending in wlr_layer_surface_v1

Sevz17:
as seen in swaywm/wlroots@59fa363


djpohly:
Cool, thanks!

PR #151, opened 2021-09-24
send frame_done also to all layer surfaces

Sevz17:
this fixes an issue when bemenu don't update his surface when typing

also apply #150


djpohly:
Thanks!

PR #152, opened 2021-09-27
remove redundancies in xcursors

ARDiDo:
`xwaylandready()` already sets the xcursor theme and size, so there is no need to do it in `setup()`. The global variables didn't appear to be used anywhere else outside of `setup()`.


djpohly:
@alex-courtis What was the reason for the second `xcursor_manager`, and is it still applicable?

Sevz17:
https://github.com/djpohly/dwl/pull/32#issuecomment-678909458

djpohly:
D'oh... thanks for pointing that out. I'm game to merge the two for now to reclaim some LOC, and if we get bug reports from XWayland users, we'll know where to start looking.

PR #158, opened 2021-10-06
add -q option to disable printing to stdout

Humm42:
Printing the status to stdout can be undesirable, as blocking pipes with
a filled buffer … block.  Because of that, when using -s for starting a
process not reading its stdin, such as service managers, after some
usage dwl will freeze.


Humm42:
> How about closing stdin on the service manager?

would work just fine; seems less intuitive to me—would still mention the problem in the ~~docs~~ README

PR #160, opened 2021-10-13
-s: close unused fds

Humm42:
dup2 doesn’t close fds, it only duplicates them.  The old ones weren’t
closed, causing problems (like dwl blocking due to the child process
never reading from the reading end, even if stdin has been closed).

I had the same problem as before—dwl blocking—and thought to myself:
Are the pipes closed properly? As it turns out, they aren’t. I believe
that’s the current problem.


djpohly:
Nice catch. Thanks for the fix!

PR #169, opened 2021-11-06

raphaelr:
Check the wlr_layer_surface_v1_state.committed bitmask to see if we need to rearrange. This is also what sway does.

Without this check, every commit request (even if only the attached buffer changed) will lead to another configure event, which will lead to another commit, etc.

This loop results in swaybg consuming 100% CPU.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -150,6 +150,7 @@ typedef struct {
 
 	struct wlr_box geo;
 	enum zwlr_layer_shell_v1_layer layer;
+	bool mapped;
 } LayerSurface;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -150,6 +150,7 @@ typedef struct {
 
 	struct wlr_box geo;
 	enum zwlr_layer_shell_v1_layer layer;
+	bool mapped;
```
Sevz17, line 153: Suckless style: `int` not `bool`

Sevz17: @raphaelr are you there?

raphaelr:
Sorry, missed this conversation.

I've just updated my build to current main, and the high CPU usage does not occur anymore. I think scenegraph fixed this, but I did not verify that. Closing this.

PR #171, opened 2021-11-13
fullscreennotify: don't crash if called before map

### LINKED ISSUE #170: dwl segfault when run something depend on libSDL2
such like: SDL_VIDEODRIVER=wayland ffplay something.mp4

raphaelr:
SDL2 calls xdg_toplevel.unset_fullscreen() before the surface is
mapped. This causes a segfault in dwl because setfullscreen() expects
the surface to be mapped already.
Therefore, delay the setfullscreen call until the surface is mapped.

Fixes #170

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1042,7 +1042,13 @@ void
 fullscreennotify(struct wl_listener *listener, void *data)
 {
 	Client *c = wl_container_of(listener, c, fullscreen);
-	setfullscreen(c, !c->isfullscreen);
+	struct wlr_xdg_toplevel_set_fullscreen_event *event = data;
+	if (!c->mon) {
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -1042,7 +1042,13 @@ void
 fullscreennotify(struct wl_listener *listener, void *data)
 {
 	Client *c = wl_container_of(listener, c, fullscreen);
-	setfullscreen(c, !c->isfullscreen);
+	struct wlr_xdg_toplevel_set_fullscreen_event *event = data;
```
Sevz17, line 1045: This function is intended to be called for both xdg and xwayland clients, so you can't use this event

Sevz17:
Merged in 03e167dbb70fbc967e310f95200bcd63f43cac72

PR #189, opened 2022-02-02
Shift+6 generates XKB_KEY_asciicircum

### LINKED ISSUE #188: SHIFT+ tag/view changes do not work on tag '6'
If a tag change or view change involving the '6' tab is initiated with a SHIFT modifier, the view or tag change does not occur as expected.

View/tag changes that don't use the SHIFT modifier work correctly.
View/tag changes to tags other than '6' work correctly.

fauxmight:
'Shift+6' generates 'XKB_KEY_asciicircum' not 'XKB_KEY_caret'.

This allows tag and view changes mapped to SHIFT[+other modifier(s)]+6 to work correctly

Resolves #188


djpohly:
Thanks for the fix!

PR #196, opened 2022-02-23
die on allocation failure

noocsharp:
While ideally we should try to recover from allocation failure, this is better than ignoring it.


Sevz17:
Thanks!

Issue #173, opened 2021-12-22
Crashes when launching a Gtk4 application


jbeich:
### Steps to reproduce
1. Start dwl
2. Start gtk4-demo (via key or from a terminal)

### Details
```c
* thread #1, name = 'dwl', stop reason = signal SIGSEGV: invalid address (fault address: 0x48)
    frame #0: 0x000000000020b8fb dwl`setfullscreen(c=0x000000083c738cc0, fullscreen=1) at dwl.c:1036:23
   1033                 c->prevy = c->geom.y;
   1034                 c->prevheight = c->geom.height;
   1035                 c->prevwidth = c->geom.width;
-> 1036                 resize(c, c->mon->m.x, c->mon->m.y, c->mon->m.width, c->mon->m.height, 0);
   1037         } else {
   1038                 /* restore previous size instead of arrange for floating windows since
   1039                  * client positions are set by the user and cannot be recalculated */
(lldb) bt
* thread #1, name = 'dwl', stop reason = signal SIGSEGV: invalid address (fault address: 0x48)
  * frame #0: 0x000000000020b8fb dwl`setfullscreen(c=0x000000083c738cc0, fullscreen=1) at dwl.c:1036:23
    frame #1: 0x000000000020b5fc dwl`fullscreennotify(listener=0x000000083c738d70, data=0x00007fffffffda60) at dwl.c:1049:2
    frame #2: 0x0000000801ee71ba libwlroots.so.10`wlr_signal_emit_safe(signal=0x000000083c73d8d0, data=0x00007fffffffda60) at signal.c:29:3
    frame #3: 0x0000000801eb73ba libwlroots.so.10`xdg_toplevel_handle_unset_fullscreen(client=0x000000081d22b700, resource=0x000000081d113e80) at wlr_xdg_toplevel.c:404:2
    frame #4: 0x0000000811bfb0dd libffi.so.7`ffi_call_unix64 at unix64.S:101
    frame #5: 0x0000000811bfa3b2 libffi.so.7`ffi_call_int(cif=0x00007fffffffdbd0, fn=(libwlroots.so.10`xdg_toplevel_handle_unset_fullscreen at wlr_xdg_toplevel.c:392), rvalue=0x0000000000000000, avalue=0x00007fffffffdc00, closure=0x0000000000000000) at ffi64.c:669:3
[...]

PR #200, opened 2022-03-14
fix overlap of gtk4 applications

Sevz17:
as mentioned in https://github.com/djpohly/dwl/issues/173#issuecomment-1000294048

TODO: `pointerfocus()` (or `buttonpress()`/`motionnotify()`?) misdetect pointer coordinates in apps


Sevz17:
Fixed with scenegraph

PR #201, opened 2022-03-15
new functions: ecalloc() and die()

Sevz17:
I don't know if this is worth, since in dwm they are used in two files (dwm.c and drw.c) but here only are used in dwl.c to save 11 lines

@djpohly what do you think?


Sevz17:
Hey @djpohly taking advantage of the fact that you show signs of life, how about this, it will allow us to compile with -pedantic

Sevz17:
Ok, later I'll merge this.

Issue #52, opened 2020-08-30
Cursor acting like it's in the top left when moved to one of the edges of the screen


Bonicgamer:
Whenever I move the cursor to the edge of the screen, it acts like the cursor is in the top left.
It's noticeable when I'm trying to click on a different tab in Firefox and then I go to the tab that's at the very left.

notthewave:
This happens when hovering over the window border, but not when the cursor is moved to the edge of the screen.

PR #204, opened 2022-03-18
merge scenegraph3

Sevz17:
It's finally ready to merge

Fixes: #52
Closes: #161 
Fixes: #75 
Closes: #143 
Fixes: #85 
Closes: #87 

Notes:
- The independent list is not used but I leave it in case we use it in the future.
- Render even if there are a client with a resize seem to works fine, however I started using dwl after d6a18d69c0c104a3f3ce3bb67d984b4479f6717e and I don't know how was before it


djpohly:
I gave it a test with my multi-monitor setup, and none of the issues I was having with the scenegraph branch before are still around. +1 from me.

If `independents` is no longer used, I'm good with removing it.  Not like it'd be hard to add back.

fauxmight:
Tested on multi-monitor proprietary nVidia desktop and on a Thinkpad T540p with the 470xx series proprietary drivers. There are issues with xwayland and cursors. There are issues with suspend/resume (with @Sevz17's idle-inhibitor branch merged) failing in a different fashion than it does with the 'main' branch, but no real regressions; nothing fails that worked prior to running this branch. Thank you for all the recent work on dwl, @Sevz17.

Frewacom:
Thanks for the good work @Sevz17, keep it up! 🥇

PR #206, opened 2022-03-23
inline the presentation variable

guidocella:
This variable can be removed since with scene-graph
wlr_presentation_surface_sampled_on_output no longer needs to be called.


Sevz17:
Good catch, thanks!

PR #210, opened 2022-03-25

ioraff:
I am still affected by the problem described in #169, so here's the PR again with the desired changes.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -150,6 +150,7 @@ typedef struct {
 	struct wl_listener surface_commit;
 
 	struct wlr_box geo;
+	int mapped;
 } LayerSurface;
 
 typedef struct {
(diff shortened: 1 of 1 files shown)


Sevz17:
Thanks, for making the changes, but I'd rather you amend @raphaelr's commit (and add yourself as co-author).

dwl.c
```diff
@@ -755,6 +756,12 @@ commitlayersurfacenotify(struct wl_listener *listener, void *data)
...
 			layers[wlr_layer_surface->current.layer]);
 
+	if (wlr_layer_surface->current.committed == 0 &&
+		layersurface->mapped == wlr_layer_surface->mapped)
+		return;
+
+	layersurface->mapped = wlr_layer_surface->mapped;
+
 	if (!wlr_output || !(m = wlr_output->data))
 		return;
```
Sevz17, line 766:
```suggestion
	if (!wlr_output || !(m = wlr_output->data))
		return;
		
	if (wlr_layer_surface->current.committed == 0
		&& layersurface->mapped == wlr_layer_surface->mapped)
	return;

	layersurface->mapped = wlr_layer_surface->mapped;
```

Issue #177, opened 2021-12-31
waitpid for Xwayland fork failed: No child processes


jbeich:
After building with `-DXWAYLAND` starting X11 apps still fails. Xwayland in other compositors works fine: sway, hikari, wayfire, cage, river, labwc, kwin_wayland, mutter, arcan-wayland.

```
$ dwl &
$ glxgears
Error: couldn't open display :0
[...]
00:00:11.336 [xwayland/server.c:92] Starting Xwayland on :0
00:00:11.448 [types/wlr_surface.c:741] New wlr_surface 0x820961a00 (res 0x82094df80)
00:00:11.451 [xwayland/server.c:243] waitpid for Xwayland fork failed: No child processes
(EE) failed to read Wayland events: Broken pipe
```

```
$ cc --version
FreeBSD clang version 13.0.0 (git@github.com:llvm/llvm-project.git llvmorg-13.0.0-0-gd7b669b3a303)
Target: x86_64-unknown-freebsd14.0
Thread model: posix
InstalledDir: /usr/bin

$ pkg info -x dwl wayland wlroots
dwl-0.2.2
wayland-1.20.0
wayland-protocols-1.24
xwayland-devel-21.0.99.1.156
wlroots-0.15.0_1
```

KNIX3:
Any update on this? I'm having the same issue on even on Linux rather than FreeBSD.

Sevz17:
The problem is caused because we setup a SIGCHLD handler.

I think we can fix it with a double fork in `spawn()`

djpohly:
Could you check to see if the recently merged #212 fixes the issue?

jbeich:
Note, merging in non-default branch is why this issue wasn't automatically closed.

jbeich:
Do you plan to merge e08bd1292288 into `main` branch? I'd like to avoid having to rebase it [downstream](https://github.com/freebsd/freebsd-ports/commit/d0ada09b16b1):
```
$ git checkout v0.3 && git cherry-pick -x e08bd1292288
Auto-merging dwl.c
[detached HEAD c7b5bb7] make sure to leave XWayland process waitable
 Author: Devin J. Pohly <djpohly@gmail.com>
 Date: Tue Mar 29 15:55:06 2022 -0500
 1 file changed, 7 insertions(+), 2 deletions(-)


$ git checkout main && git cherry-pick -x e08bd1292288
Auto-merging dwl.c
CONFLICT (content): Merge conflict in dwl.c
[...]

PR #211, opened 2022-03-29
double forking for process spawning

Sevz17:
Fixes: #177

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1748,22 +1748,31 @@ run(char *startup_cmd)
 
 	/* Now that the socket exists, run the startup command */
 	if (startup_cmd) {
+		/* For explanation about double-forking see spawn() */
 		int piperw[2];
 		pipe(piperw);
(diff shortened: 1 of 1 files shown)


djpohly:
I really don't like double-forking.  It leaves orphaned processes and messes up the process tree.

Sevz17:
Try wlroots-next, Xwayland won't be able to spawn

Sevz17:
> It leaves orphaned processes and messes up the process tree.

We could do something like [this](https://github.com/swaywm/sway/blob/master/sway/commands/exec_always.c#L48-L92) to avoid orphaned processes

djpohly:
Well, I'm still bummed we can't just SIG_IGN the whole thing because of the XWayland fork/wait, but I put together an idea over in #212 which seems to work (and as a bonus, is shorter :grin:).

Sevz17:
Closed in favor of #212

PR #213, opened 2022-03-31
natural_scrolling is ignored

dimkr:


djpohly:
I'm not sure I see how this improves things, since `natural_scrolling` is passed to the libinput function to enable or disable it as desired. This patch actually appears to make it impossible to *dis*able natural scrolling if it's enabled by default.  Am I missing a detail here?

PR #214, opened 2022-03-31
Fall back to a lower output mode if needed (swaywm/sway@4cdc4ac)

dimkr:
I have a Panasonic TV with an unsupported preferred mode. It works only when the compositor falls back to a lower resolution.


djpohly:
Is this something that could best be solved by adding a quirk to the kernel's drm_edid.c that fixes the preferred mode for everyone?

PR #215, opened 2022-03-31
Add basic support for pointer constraints

dimkr:
Without this, mouse movement in FPS games is chaotic. See https://github.com/puppylinux-woof-CE/woof-CE/issues/2534 and https://github.com/puppylinux-woof-CE/woof-CE/issues/2593. 

This is a simplified version of johanmalm/labwc#85.

Reproduced and tested using Crispy Doom, running both natively (with free look enabled) and under Xwayland. Also tested QEMU mouse grab and ungrab under Xwayland.

I'm not sure if I got everything right, especially the handling of destroy_constraint, and the `if constraint` check on input events.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,9 +56,19 @@ idle-protocol.c:
 
 idle-protocol.o: idle-protocol.h
 
+pointer-constraints-unstable-v1-protocol.h:
+	$(WAYLAND_SCANNER) server-header \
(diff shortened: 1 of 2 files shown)


PalanixYT:
There is a patch for that already. See https://github.com/djpohly/dwl/wiki/Patches

PR #216, opened 2022-03-31

ioraff:
Carried over from #169, with the original commit amended.

> Check the wlr_layer_surface_v1_state.committed bitmask to see if we need
> to rearrange. This is also what sway does.
> 
> Without this check, every commit request (even if only the attached buffer
> changed) will lead to another configure event, which will lead to another
> commit, etc.
> 
> This loop results in swaybg consuming 100% CPU.


Sevz17:
Thanks!

PR #217, opened 2022-03-31
drop permissions

ioraff:
Resurrecting #36 with some style changes.

This works for me given `seatd` built with `-Dlibseat-builtin=enabled`, `chown :input $(DESTDIR)$(PREFIX)/bin/dwl`, and
`chmod g+s    $(DESTDIR)$(PREFIX)/bin/dwl`. My user is NOT in the `input` group, but is in the `audio`, `video`, and `tty` groups.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1117,6 +1117,23 @@ dirtomon(enum wlr_direction dir)
 	return selmon;
 }
 
+static void
+drop_permissions(void)
+{
+	if (getuid() != geteuid() || getgid() != getegid()) {
+		/* Set the gid and uid in the correct order. */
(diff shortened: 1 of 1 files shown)


djpohly:
What exactly is the use case for running dwl as root?  I don't think I quite understood it the first time either.

ioraff:
It eliminates the need to run a daemon like (e)logind or seatd as root, which is overkill on a single-user system.

Sevz17:
Sway deprecated SUID (https://github.com/swaywm/sway/commit/e5728052b59fb5b476c78c9f18b812a85d7f4503)

Issue #125, opened 2021-05-28
Steam menus drawn in wrong position


TRPB:
When running Steam, the application works but when it starts up it's not drawn as a full tile until the tiles are rearranged somehow (e.g. adding a new tile). This isn't really a problem on its own since it's very easily fixed.

However, the menus seem drawn arbitrarily. In the state mentioned above, I have steam running on display 1 and the drop down menus (store, community, etc) appear on display 2. Once the tiles have been rearranged and it's drawn correctly, Steam's menus are drawn about 300px left of where they should be.

adigitoleo:
I've seen this happen in various GUI programs as well, I haven't tracked it down properly yet but I think I remember it happening more (only?) when the GUI is running under XWayland. Do you know if Steam runs natively on wayland, or does it use XWayland? You can check using `xprop` (click on the window, if it prints something, then it's XWayland).

pm4rcin:
@adigitoleo Steam of course runs on Xwaylad.

Sevz17:
@TRPB, is this still happening?

pm4rcin:
@Sevz17 I think I have the same problem. I have no external monitors. When I play e.g. Portal 2 while steam window is open on another tag when I click in game it registers clicks in steam and not in game. Another example when I have freetube (XWayland like steam) open on e.g. tag 2 mouse clicks keepassxc menus on another tab. It's reproducible every time.

HadetTheUndying:
This is because Steam menus spawn a fresh Window(XWayland windows) for every menu and submenu some ungodly reason.

Sevz17:
> @Sevz17 I think I have the same problem. I have no external monitors. When I play e.g. Portal 2 while steam window is open on another tag when I click in game it registers clicks in steam and not in game. Another example when I have freetube (XWayland like steam) open on e.g. tag 2 mouse clicks keepassxc menus on another tab. It's reproducible every time.

I think that's a different bug

pm4rcin:
@Sevz17 yeah probably yes. How could I debug it or do you have any idea what could be the cause?

Sevz17:
@pm4rcin, maybe related: Yesterday I realized I can't click in the settings window of VirtualBox if it is overlapping its parent.
Although I recommend open another issue for this.

Sevz17:
Can anyone tell me if #218 fixes this issue?

fauxmight:
[...]

PR #218, opened 2022-04-01
arrange client's monitor if size has change since last commit

sevz17:
Fixes: #125


fauxmight:
This is working well and I have regularly built with it for the two months since it was posted here.
Any reason this cannot be pushed to mainline?

PR #221, opened 2022-04-02
Remove unused 'x' and 'y' fields from MonitorRule struct

### LINKED ISSUE #220: MonitorRule 'x' and 'y' elements are never used; monitors cannot be consistently positioned
## Info
dwl's commit: 4d3adea68325b70eafbd56a31023f96b92532b8d
wlroots version: 0.15.1
## Description

Elements 'x' and 'y' in the MonitorRule struct do not appear to ever be used.
It is not clear whether 'x' and 'y' were originally intended to indicate _position_  or _resolution_ of a monitor, but the claim in `config.def.h` that

 `The order in which monitors are defined determines their position.`

does not bear out to be true. Despite ordering three outputs (monitors) in A, B, C order in `config.h`, they are laid out in A, C, B order in reality. Changing 'x' and 'y' structure elements to match either desired position or resolution of the monitors has no effect.

fauxmight:
This closes #220

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -29,12 +29,12 @@ static const Layout layouts[] = {
  * The order in which monitors are defined determines their position.
  * Non-configured monitors are always added to the left. */
 static const MonitorRule monrules[] = {
-	/* name       mfact nmaster scale layout       rotate/reflect x y */
+	/* name       mfact nmaster scale layout       rotate/reflect */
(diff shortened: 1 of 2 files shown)


fauxmight:
Closing this pull request. The 'x' and 'y' fields are not currently in use but may be in the future and probably should be.

Discussed further in #222

PR #223, opened 2022-04-03
restack xwayland surface

Sevz17:


fauxmight:
Major problems with this:
Prior to merging this branch, I was using dwl very efficiently as starting Steam games proved to be nearly impossible.

Now, I can play X3 and I am not sure I will get anything else accomplished.
: )

pm4rcin:
The problem is that both Steam and keepassxc use X11. But I was able to install qt5-wayland so the problem is... gone (obviously). But the simplest test is to just use xeyes and see for yourself that Steam menus are rendered on tag where it (X11 client) is present (except that xeyes doesn't have any menu). 
If you spawn Keepass and then Steam it registers clicks on Steam.
If you spawn Steam and then Keepass it registers clicks on Keepass.
I've opened Signal-Desktop (still X11, wayland coming very soon) on another tag and every single XWayland Window registers clicks on Signal.
I've moved Signal to keepass on tag 4 so it's in tiling. Signal on right, steam on left side. Keypass on tag 3 registers clicks on Signal but only on the right side, the left side works fine. Steam registers clicks from keepassxc since keepassxc was spawned after steam and keepassxc occupies whole tag 3.
In summary: Whatever X11 client you spawn last it behaves like being on top of every other XWayland window. Every X11 client spawns on top of previous no matter which tag it is on.
PS. Also a bit off-topic but every XWayland window spawns as floating by default (is it intended?).

BenJarg:
@pm4rcin I was able to solve the problem with X windows registering clicks on one another by adding the same 5 lines of this PR at the end of the "focusclient" function. I'm not sure if this is a reasonable thing to do, or a hack, but it fixed things for me nonetheless.

Sevz17:
@djpohly, do you know if this will have any undesired effects?

pm4rcin:
I have tested it today. Looks like there are no side effects.

PR #226, opened 2022-04-09
implement ext-session-lock-v1

sevz17:
Implement ext-session-lock which is merged in [wayland-protocols](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/merge_requests/131)

Currently, our security policy is to allow creating new screenlockers if the previous one dies, but if the previous screenlocker dies and no new screenlocker is started, dwl will stay locked.

PS. I don't like this implementation, but wlroots' one requires us to unlink all our handlers.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -40,6 +40,7 @@
 #include <wlr/types/wlr_screencopy_v1.h>
 #include <wlr/types/wlr_seat.h>
 #include <wlr/types/wlr_server_decoration.h>
+#include <wlr/types/wlr_session_lock_v1.h>
 #include <wlr/types/wlr_single_pixel_buffer_v1.h>
 #include <wlr/types/wlr_subcompositor.h>
 #include <wlr/types/wlr_viewporter.h>
@@ -73,7 +74,7 @@
 /* enums */
 enum { CurNormal, CurPressed, CurMove, CurResize }; /* cursor */
 enum { XDGShell, LayerShell, X11Managed, X11Unmanaged }; /* client types */
-enum { LyrBg, LyrBottom, LyrTop, LyrOverlay, LyrTile, LyrFloat, LyrFS, LyrDragIcon, NUM_LAYERS }; /* scene layers */
+enum { LyrBg, LyrBottom, LyrTop, LyrOverlay, LyrTile, LyrFloat, LyrFS, LyrDragIcon, LyrBlock, NUM_LAYERS }; /* scene layers */
 #ifdef XWAYLAND
 enum { NetWMWindowTypeDialog, NetWMWindowTypeSplash, NetWMWindowTypeToolbar,
 	NetWMWindowTypeUtility, NetLast }; /* EWMH atoms */
@@ -173,6 +174,8 @@ struct Monitor {
 	struct wlr_scene_rect *fullscreen_bg; /* See createmon() for info */
 	struct wl_listener frame;
 	struct wl_listener destroy;
+	struct wl_listener destroy_lock_surface;
+	struct wlr_session_lock_surface_v1 *lock_surface;
 	struct wlr_box m;      /* monitor area, layout-relative */
 	struct wlr_box w;      /* window area, layout-relative */
 	struct wl_list layers[4]; /* LayerSurface::link */
@@ -202,6 +205,15 @@ typedef struct {
 	int monitor;
 } Rule;
 
+typedef struct {
+	struct wlr_scene_tree *scene;
+
+	struct wlr_session_lock_v1 *lock;
+	struct wl_listener new_surface;
+	struct wl_listener unlock;
+	struct wl_listener destroy;
+} SessionLock;
+
 /* function declarations */
 static void applybounds(Client *c, struct wlr_box *bbox);
 static void applyrules(Client *c);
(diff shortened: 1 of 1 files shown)


djpohly:
Would this obviate the use-case for input-inhibit?

Sevz17:
I think so, this is within wayland-protocols, while input-inhibit is a wlroots-protocol, however input-inhibit only added like 4 lines

sevz17:
@fauxmight, when you upgrade to wlroots-0.16 can you try this?, I'm interested in plug/unplug and enable/disable monitors.

fauxmight:
I don't have any difference with multi-monitor connect/disconnect with or without the "ext-session-lock-v1" branch merged:

(Starting with two monitors connected and active):
When one monitor is disconnected, clients are moved (correctly) to the remaining, connected monitor.

When the second monitor is disconnected [wait 10 seconds] and reconnected,
- no segfault happens immediately (as it did with 0.15x)
- no mouse/keyboard interaction with any client seems possible
- attempting to MOVE a client window results in a dwl segfault

sevz17:
> * attempting to MOVE a client window results in a dwl segfault

Can you describe the steps to reproduce the segfault?

fauxmight:
Thanks for today's updates (2022-12-01). This makes testing a bit easier.

Environment includes
```
export MOZ_ENABLE_WAYLAND=1
export GDK_BACKEND=wayland
```
Testing with the following clients: alacritty and firefox and (layer shell) waybar

Start dwl with two active monitors
Start alacritty on monitor1
Start firefox on monitor2
Disconnect monitor2 with KVM switch
[At this point client on monitor2 is moved to monitor1 retaining its tags and all continues working/behaving correctly.]
Power-off monitor1
[Now, no active monitor remains]
Wait 10 seconds
Power-on monitor1
At this point with all monitors in tile mode, alacritty is resized to fill an entire monitor no longer honoring the layer shell client space. If alacritty is set transparent beforehand, waybar still shows through the transparent alacritty from underneath (demonstrating that waybar has not been killed in the monitor disconnects/reconnects).

Shortly after showing this way, dwl dies.

Log shows the following:

```
amdgpu: amdgpu_cs_ctx_create2 failed. (-13)
Gdk-Message: 13:15:54.945: Error reading events from display: Broken pipe
Error: "Event loop terminated with code: 1"
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
Exiting due to channel error.
```

sevz17:
> I don't have any difference with multi-monitor connect/disconnect with or without the "ext-session-lock-v1" branch merged:

Ah, I missed this one, I guess I can merge this and fix the crash in main, right?

sevz17:
> Thanks for today's updates (2022-12-01). This makes testing a bit easier.
> 
> Environment includes
> 
> ```
> export MOZ_ENABLE_WAYLAND=1
> export GDK_BACKEND=wayland
> ```
> 
> Testing with the following clients: alacritty and firefox and (layer shell) waybar
> 
> Start dwl with two active monitors Start alacritty on monitor1 Start firefox on monitor2 Disconnect monitor2 with KVM switch [At this point client on monitor2 is moved to monitor1 retaining its tags and all continues working/behaving correctly.] Power-off monitor1 [Now, no active monitor remains] Wait 10 seconds Reconnect monitor2 with KVM switch At this point with all monitors in tile mode, alacritty is resized to fill an entire monitor no longer honoring the layer shell client space. If alacritty is set transparent beforehand, waybar and firefox still show through the transparent alacritty from underneath (demonstrating that neither waybar nor firefox has not been killed in the monitor disconnects/reconnects).
> 
> Attempting to change tags on alacritty causes dwl to die. Attempting to move alacritty by mouse causes dwl to die.
> 
> Log shows the following:
> 
> ```
> amdgpu: amdgpu_cs_ctx_create2 failed. (-13)
> Gdk-Message: 13:15:54.945: Error reading events from display: Broken pipe
> Error: "Event loop terminated with code: 1"
> Exiting due to channel error.
> Exiting due to channel error.
> Exiting due to channel error.
> Exiting due to channel error.
> Exiting due to channel error.
> ```
> 

Can you send me the backtrace?

>     1. Will try with something other than alacritty to make sure it is not specific to alacritty.

Thanks!

>     2. Considering the log above, this may be an issue specific to AMD GPU.

From a time ago I started to see this `amdgpu: amdgpu_cs_ctx_create2 failed. (-13)` when starting dwl/sway, but I haven't experienced issues related to it (but I run a single-monitor setup, things may change)

sevz17:
@fauxmight?

fauxmight:
Will test soon. Apologies for the delay.

sevz17:
> Will test EDIT ~soon~ within two days. Apologies for the delay.

Ok, no rush, thank you for helping me to test

fauxmight:
> > I don't have any difference with multi-monitor connect/disconnect with or without the "ext-session-lock-v1" branch merged:
> 
> Ah, I missed this one, I guess I can merge this and fix the crash in main, right?

I cannot see the crash in main that this fixes, but this still makes no problems for me with single- or dual-monitor. As far as I can tell, this is good to merge.

fauxmight:
Thank you, @sevz17 for the amazing quantity of work you have put into dwl, especially in recent days.

PR #228, opened 2022-04-25
Run printstatus() when a monitor is removed

BenJarg:
Windows on the closed monitor will be sent to another monitor, so the status of occupied tags, etc. should be updated to reflect this for external bars.


Sevz17:
Thanks!

Issue #141, opened 2021-08-15
Better Touchpad Support


shadow5013:
A lot of features (e.g. disable while typing,  acceleration, middle click emulation, tap and drag ) are missing. It would be nice to have them.

xi:
Can you give more context? As far as I understand, most of this can already be achieved by patching `createpointer()`. See also the [libinput docs on configuration](https://wayland.freedesktop.org/libinput/doc/latest/configuration.html). If by "tab and drag" you mean drag and drop see #134 and #144.

Does this answer your question or is there something missing?

shadow5013:
I kinda expected to configure touchpad from the `config.h` much like keyboard.

pm4rcin:
@dimkr as mentioned [here](https://github.com/djpohly/dwl/pull/229#issuecomment-1129107612) we want it to be setup by config.h so if you could rework your PR it'd be great.

dimkr:
> @dimkr as mentioned https://github.com/djpohly/dwl/pull/229#issuecomment-1129107612 we want it to be setup by config.h so if you could rework your PR it'd be great.

My use case is different: I want to be able to change these settings without recompiling, because it's a binary used by multiple users (some are left-handed, etc').

PR #229, opened 2022-04-26
Add configuration options for touchpads

BenJarg:
This may make more sense as a patch, but seems like a somewhat basic configuration option for touchpads (and has been asked for in #141).

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -49,7 +49,31 @@ static const int repeat_delay = 600;
 
 /* Trackpad */
 static const int tap_to_click = 1;
+static const int tap_and_drag = 1;
+static const int drag_lock = 1;
 static const int natural_scrolling = 0;
+static const int disable_while_typing = 1;
+static const int left_handed = 0;
+static const int middle_button_emulation = 0;
+/* You can choose between:
+LIBINPUT_CONFIG_SCROLL_NO_SCROLL
+LIBINPUT_CONFIG_SCROLL_2FG
+LIBINPUT_CONFIG_SCROLL_EDGE
+LIBINPUT_CONFIG_SCROLL_ON_BUTTON_DOWN
+*/
+static const enum libinput_config_scroll_method scroll_method = LIBINPUT_CONFIG_SCROLL_2FG;
+/* You can choose between:
+LIBINPUT_CONFIG_SEND_EVENTS_ENABLED
+LIBINPUT_CONFIG_SEND_EVENTS_DISABLED
+LIBINPUT_CONFIG_SEND_EVENTS_DISABLED_ON_EXTERNAL_MOUSE
+*/
+static const uint32_t send_events_mode = LIBINPUT_CONFIG_SEND_EVENTS_ENABLED;
+/* You can choose between:
+LIBINPUT_CONFIG_ACCEL_PROFILE_FLAT
+LIBINPUT_CONFIG_ACCEL_PROFILE_ADAPTIVE
+*/
+static const enum libinput_config_accel_profile accel_profile = LIBINPUT_CONFIG_ACCEL_PROFILE_ADAPTIVE;
+static const double accel_speed = 0.0;
 
 /* If you want to use the windows key change this to WLR_MODIFIER_LOGO */
 #define MODKEY WLR_MODIFIER_ALT
--- a/dwl.c
+++ b/dwl.c
@@ -981,17 +981,36 @@ createpointer(struct wlr_input_device *device)
 		struct libinput_device *libinput_device =  (struct libinput_device*)
 			wlr_libinput_get_device_handle(device);
 
-		if (tap_to_click && libinput_device_config_tap_get_finger_count(libinput_device))
-			libinput_device_config_tap_set_enabled(libinput_device, LIBINPUT_CONFIG_TAP_ENABLED);
+		if (libinput_device_config_tap_get_finger_count(libinput_device)) {
+			libinput_device_config_tap_set_enabled(libinput_device, tap_to_click);
+			libinput_device_config_tap_set_drag_enabled(libinput_device, tap_and_drag);
+			libinput_device_config_tap_set_enabled(libinput_device, drag_lock);
+		}
 
 		if (libinput_device_config_scroll_has_natural_scroll(libinput_device))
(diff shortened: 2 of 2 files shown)


Sevz17:
It may worth adding more options, like changing the default accel speed, and also the default accel profile

Sevz17:
It will definitely be by config.h

BenJarg:
So now this configures disable while typing, the accel speed, and accel profile. Is there anything else reasonable that should be on this list?

pm4rcin:
[Here's](https://wayland.freedesktop.org/libinput/doc/latest/configuration.html) all the available options. @Sevz17 do we want every option from there?

BenJarg:
Thanks, perhaps left handed mode and middle button emulation should also be added. I suppose we should decide what exactly to include though, and whether or not we want everything available.

config.def.h
```diff
@@ -50,6 +50,9 @@ static const int repeat_delay = 600;
 /* Trackpad */
 static const int tap_to_click = 1;
 static const int natural_scrolling = 0;
+static const int disable_while_typing = 1;
+static const enum libinput_config_accel_profile accel_profile = LIBINPUT_CONFIG_ACCEL_PROFILE_ADAPTIVE;
```
Sevz17, line 54: add a comment like:
```
You can choose between
LIBINPUT_CONFIG_ACCEL_PROFILE_FLAT
LIBINPUT_CONFIG_ACCEL_PROFILE_ADAPTIVE
```

Sevz17:
- tap to click
- tap and drag
- drag lock
- left handled mode
- middle button emulation
- [libinput_device_config_scroll_set_method()](https://wayland.freedesktop.org/libinput/doc/latest/api/group__config.html#ga482951a2b1faf837e43d30d9c92dc9d3)
- [libinput_device_config_send_events_set_mode()](https://wayland.freedesktop.org/libinput/doc/latest/api/group__config.html#gaacc79635146b888ddaf1cb35363b22ce)
- [libinput_device_config_dwtp_set_enabled()](https://wayland.freedesktop.org/libinput/doc/latest/api/group__config.html#ga32a8b64126f3a03aabaab367a7dadc98)

I think these options should be configured, does anyone want to add or remove some of these?

pm4rcin:
Basically we will have all the options for Touchpad so I think it's good. I think we should provide comments for every option so people don't have to look for docs etc.

BenJarg:
Ok, I've added everything on that list (except dwtp), and comments with values for the options of things that you would need to look up. I didn't added dwtp because it looks like it won't be available until libinput 1.21, and the latest release is 1.20, so as far as I can tell it's a future setting (someone correct me if I misunderstood the docs).

dwl.c
```diff
@@ -981,11 +981,50 @@ createpointer(struct wlr_input_device *device)
...
-		if (tap_to_click && libinput_device_config_tap_get_finger_count(libinput_device))
-			libinput_device_config_tap_set_enabled(libinput_device, LIBINPUT_CONFIG_TAP_ENABLED);
+		if (libinput_device_config_tap_get_finger_count(libinput_device)) {
+			if (tap_to_click)
+				libinput_device_config_tap_set_enabled(libinput_device, LIBINPUT_CONFIG_TAP_ENABLED);
+
+			if (tap_and_drag)
+				libinput_device_config_tap_set_drag_enabled(libinput_device, LIBINPUT_CONFIG_DRAG_ENABLED);
+			else
+				libinput_device_config_tap_set_drag_enabled(libinput_device, LIBINPUT_CONFIG_DRAG_DISABLED);
```
Sevz17, line 991: I think is better just `libinput_device_config_tap_set_drag_enabled(libinput_device, tap_and_drag);`

dwl.c
```diff
@@ -981,11 +981,50 @@ createpointer(struct wlr_input_device *device)
...
+
+			if (drag_lock)
+				libinput_device_config_tap_set_enabled(libinput_device, LIBINPUT_CONFIG_TAP_ENABLED);
+			else
+				libinput_device_config_tap_set_enabled(libinput_device, LIBINPUT_CONFIG_TAP_DISABLED);
```
Sevz17, line 996: The same as above.

dwl.c
```diff
@@ -981,11 +981,50 @@ createpointer(struct wlr_input_device *device)
...
+			libinput_device_config_dwt_set_enabled(libinput_device, disable_while_typing);
+
+		if (libinput_device_config_left_handed_is_available(libinput_device))
+			libinput_device_config_left_handed_set(libinput_device, left_handed);
+
+		if (libinput_device_config_middle_emulation_is_available(libinput_device)) {
+			if (middle_button_emulation)
+				libinput_device_config_middle_emulation_set_enabled(libinput_device, LIBINPUT_CONFIG_MIDDLE_EMULATION_ENABLED);
+			else
+				libinput_device_config_middle_emulation_set_enabled(libinput_device, LIBINPUT_CONFIG_MIDDLE_EMULATION_DISABLED);
```
Sevz17, line 1012: The same as above

Sevz17:
Other thing, at least for me it's more understandable place the variables below the comment about it.

dwl.c
```diff
@@ -981,11 +981,50 @@ createpointer(struct wlr_input_device *device)
...
+		}
+
+		if (libinput_device_config_left_handed_is_available(libinput_device))
+			libinput_device_config_left_handed_set(libinput_device, left_handed);
```
Sevz17, line 1016: This is called above.

dwl.c
BenJarg: I wasn't sure if `LIBINPUT_CONFIG_DRAG_ENABLED` was guaranteed to be 1, etc. But if that's the case, you're right, this is much cleaner.  Perhaps the if statement for `tap_to_click` should be replaced with a similar line if `LIBINPUT_CONFIG_TAP_ENABLED` is always going to be 1?

Sevz17:
> Perhaps the if statement for `tap_to_click` should be replaced with a similar line if `LIBINPUT_CONFIG_TAP_ENABLED` is always going to be 1?

Yes, do it.
Also, the comment at the end of the function should be replaced.

Sevz17:
Thanks!

PR #233, opened 2022-05-11
Raise focused client scene to top when switching to monocle layout

BenJarg:
When switching from tile layout to monocle layout, sometimes the window that was focused before ends up under other windows (although still considered focused). If there is a better way to fix this, do tell me, but this does seem to fix the issue.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1405,6 +1405,7 @@ monocle(Monitor *m)
 			continue;
 		resize(c, m->w.x, m->w.y, m->w.width, m->w.height, 0);
 	}
+	focusclient(focustop(m), 1);
 }
 
 void


BenJarg:
It may also be worth noting that this is an issue with the configuration option "sloppyfocus" enabled. If I do a mouse click on the focused window also before switching to monocle layout, it then does seem to consistently end up on top. But this is not the case if I focus it only by hovering the mouse over it.

dwl.c
```diff
@@ -1398,13 +1398,15 @@ mapnotify(struct wl_listener *listener, void *data)
...
 {
-	Client *c;
+	Client *c, *sel;
 
 	wl_list_for_each(c, &clients, link) {
 		if (!VISIBLEON(c, m) || c->isfloating || c->isfullscreen)
 			continue;
 		resize(c, m->w.x, m->w.y, m->w.width, m->w.height, 0);
 	}
+	sel = selclient();
```
Sevz17, line 1408: just reuse `c`

Sevz17: also, it will segfault if there are no clients

dwl.c
```diff
@@ -1398,13 +1398,15 @@ mapnotify(struct wl_listener *listener, void *data)
...
+	wlr_scene_node_raise_to_top(sel->scene);
```
Sevz17, line 1409:
```suggestion
	focusclient(focustop(m), 1);
```

BenJarg: Right, I'm testing this now and it seems to be the better solution.

Sevz17:
Thanks!

PR #236, opened 2022-05-15
Check if XWayland client size_hints are NULL

BenJarg:
Certain XWayland clients segfault dwl because size_hints is NULL, such as the normal X version of rofi which I've accidentally run  before in a dwl session. I don't know if setting as the default sizes to 1 is necessarily the right thing to do though.

### DIFF
--- a/client.h
+++ b/client.h
@@ -215,8 +215,13 @@ client_min_size(Client *c, int *width, int *height)
 	if (client_is_x11(c)) {
 		struct wlr_xwayland_surface_size_hints *size_hints;
 		size_hints = c->surface.xwayland->size_hints;
-		*width = size_hints->min_width;
-		*height = size_hints->min_height;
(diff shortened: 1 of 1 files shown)


client.h
```diff
@@ -215,8 +215,14 @@ client_min_size(Client *c, int *width, int *height)
...
 		size_hints = c->surface.xwayland->size_hints;
-		*width = size_hints->min_width;
-		*height = size_hints->min_height;
+		if (size_hints) {
+			*width = size_hints->min_width;
+			*height = size_hints->min_height;
+		}
+		else {
+			*width = 1;
+			*height = 1;
```
Sevz17, line 224:
```suggestion
		} else {
			*width = 0;
			*height = 0;
```

Sevz17:
Thanks!

PR #238, opened 2022-05-18
allow unmanaged clients (like dzen or dmenu) to have focus

sevz17:


Sevz17:
Hey @fauxmight, can you try this and tell me how it works on multi-monitor setup?

fauxmight:
I cannot tell any difference on single or on multi-monitor.
What are you doing to test the patch's effect?

Sevz17:
> I cannot tell any difference on single or on multi-monitor. What are you doing to test the patch's effect?

Sorry, yesterday when I asked it made sense, but now that I've rested and can think better, there are no differences.

fauxmight:
I cannot tell what this should change.
Is the intended effect specific to xorg applications?

Sevz17:
> I cannot tell what this should change. Is the intended effect specific to xorg applications?

Yes, this is only for xwayland clients, basically menus for X11, which we handle as unmanaged clients.

> I cannot see any change in focus behaviour with unmanaged native wayland clients like waybar or bemenu.

This is not for they, and actually they are managed as layer-shell

fauxmight:
I missed your answer from 20 May until yesterday (16 June, 2022).
OK, understanding now what this is intended to do, I am not in front of a multi-monitor setup right now, but I notice that on a single monitor setup, after clicking on a dmenu or dzen2 bar, (and the bar gaining focus), I can no longer focus any other application (whether by mouse hover or by actually clicking an outside application) until starting a new application or closing dzen2/dmenu.

This may be intended behaviour for an unmanaged client, but does not seem correct.
I will get multi-monitor testing as soon as possible.

Sevz17:
My intention is implement the same behavior than layer shell, so let me know if something is not like it

fauxmight:
This works reasonably in a multi-monitor setup and operates as layer-shell is expected to.

One oddity: if I start dmen_run from the primary monitor, it runs on the primary monitor.
If I start dmenu_run while focus is on any other monitor (for instance, starting dmenu_run from within an alacritty terminal on a secondary monitor), dmenu STILL starts on the primary monitor.

There is nothing necessarily wrong with this, but I don't know whether or not it is intended behavior.

Sevz17:
> One oddity: if I start dmen_run from the primary monitor, it runs on the primary monitor. If I start dmenu_run while focus is on any other monitor (for instance, starting dmenu_run from within an alacritty terminal on a secondary monitor), dmenu STILL starts on the primary monitor.

So, A layer surface starts on the selected monitor, doesn't it?
And also what happens if you have a layer surface open on the primary monitor and you switch to a secondary monitor?


> At present, this branch conflicts with your `arrange-commit` branch. I regularly use that branch, so I won't be doing more testing with this one unless you specifically ask for more testing.

Ok, no problem, thanks for helping me.

fauxmight:
> So, A layer surface starts on the selected monitor, doesn't it? And also what happens if you have a layer surface open on the primary monitor and you switch to a secondary monitor?

"Primary monitor" was a poor choice of words on my part, but we will continue using "primary monitor" to indicate "the monitor that is at position 0, 0."

The layer surface starts on the primary monitor no matter what.
It stays on the primary monitor no matter what.

No matter what monitor had focus when the layer surface unmanaged client was started, that unamanged client ALWAYS renders on the monitor that is at position 0, 0.
If after the unmanaged client is started, mouse focus is moved  to another monitor, the layer surface unmanaged client stays on the the monitor that is at position 0, 0.

Sevz17:
> No matter what monitor had focus when the layer surface unmanaged client was started, that unamanged client ALWAYS renders on the monitor that is at position 0, 0. If after the unmanaged client is started, mouse focus is moved to another monitor, the layer surface unmanaged client stays on the the monitor that is at position 0, 0.

I am a bit confused, for layer surface I am referring to native wayland clients (e.g. fuzzel or bemenu with wayland backend) and with unmanaged client I am referring to xwayland clients (e.g. dmenu or dzen). Can you re-explain with the previous terms?

Sevz17:
Probably we want to merge this after #259.

fauxmight:
@Sevz17: You're right. That was a very ambiguous wording. I am referring EXCLUSIVELY to xwayland unamanged clients throughout this description.

The xwayland unmanaged client starts on the primary monitor no matter what.
It stays on the primary monitor no matter what.

No matter what monitor had focus when the xwayland unmanaged client was started, that xwayland unamanged client ALWAYS renders on the monitor that is at position 0, 0.
If after the xwayland unmanaged client is started, mouse focus is moved to another monitor, the layer surface unmanaged client stays on the the monitor that is at position 0, 0.

I do NOT see any problems with the way things are working, and YES, after #259 is merged, this is functioning reasonably and can also be merged. Should you or another user feel that xwayland unmanaged clients SHOULD be startable on other monitors, that change can be made at a later date if necessary.

sevz17:
Closing this: causes other issues that requires more work (and lines) that is worth for something X11 related
If anyone really wants to work on this, feel free to fork my branch (I won't delete it)

NOTE: I found a bug, in winecfg when you click the menu to select windows version, it closes immediately.

sevz17:
I was reading the wlroots headers and realize that [wlr_xwayland_icccm_input_model()](https://gitlab.freedesktop.org/wlroots/wlroots/-/blob/master/include/wlr/xwayland.h#L317-L318) may solve the bug I mentioned w/o add too much code.

sevz17:
@fauxmight, please can you test this?, I think it's ok to merge it into main

fauxmight:
@sevz17: This is working correctly.
No problems. Looks good for merge.

sevz17:
Thanks!

PR #247, opened 2022-06-02
initialize xkb_rules.options to a null pointer

Humm42:
Initializing it to an empty string had broken configuring xkbcommon
through the environment (XKB_DEFAULT_OPTIONS).

Fixes: ae313911153b ("initialize rules and xkb_rules")

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -41,7 +41,7 @@ static const struct xkb_rule_names xkb_rules = {
 	/* example:
 	.options = "ctrl:nocaps",
 	*/
-	.options = "",
+	.options = NULL,
 };
 
 static const int repeat_rate = 25;


Sevz17:
Maybe it's worth initialize all xbk_rules with a NULL pointer, doesn't it?

djpohly:
It was initialized to a NULL pointer before. Any fields not specified in an initializer (when an initializer is given) are by C standard initialized to the zero value of their type.

Sevz17:
Well, with #245 there are no default flags, and we can leave `rules` and `xkb_rules` empty

Humm42:
>It was initialized to a NULL pointer before. Any fields not specified
>in an initializer (when an initializer is given) are by C standard
>initialized to the zero value of their type.

No, the initializer had been empty.  You can’t do that.

Other ways to initialize `xkb_rules` to all-zero:

	/* maybe here a comment about what you can put in there */
	static const struct xkb_rule_names xkb_rules; /* static storage duration, thus zero-intialized */

	static const struct xkb_rule_names xkb_rules = {
		/* can specify fields: rules, model etc. */
		/* example :
		.options = "ctrl:nocaps",
		*/
		0
	};

PR #251, opened 2022-06-12
Fix make install while files are in use

gwnichol:
When the dwl executable is in use, `cp` fails with the below error:
```
mkdir -p /usr/local/bin
cp dwl /usr/local/bin
cp: cannot create regular file '/usr/local/bin/dwl': Text file busy
make: *** [Makefile:65: install] Error 1
```

The `-f` flag creates a new file rather than using `open` on the previous file. This way the `dwl` process can still use the original file until it closes it. This appears to be the same behavior used by the `install` command.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,10 +56,10 @@ dist: clean
 
 install: dwl
 	mkdir -p $(DESTDIR)$(PREFIX)/bin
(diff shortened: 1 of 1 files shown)


Sevz17:
Thanks!

PR #259, opened 2022-06-20
do not focus clients if a layer surface is focused

sevz17:
This fix an issue when you have a layer surface mapped (e.g. fuzzel) and you spawn various clients. You can use mod+j/mod+k to change its border color but not get real focus (the same happens with mouse input). The only surface should be focused is the layer surface

Do not merging until real world testing.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -299,6 +299,7 @@ static void zoom(const Arg *arg);
 /* variables */
 static const char broken[] = "broken";
 static pid_t child_pid = -1;
+static struct wlr_surface *exclusive_focus;
 static struct wl_display *dpy;
 static struct wlr_backend *backend;
 static struct wlr_scene *scene;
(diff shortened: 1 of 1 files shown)


fauxmight:
BEFORE committing this: popups from layer surface *do* appear, but do *not* show over XDG shell clients.
Example: hovering over waybar (layer shell) elements with *no* XDG shell clients on-screen *will* allow popups from waybar to appear.
Hovering over waybar (layer shell) elements when XDG shell clients *are* on screen -> popups are hidden by XDG shell clients.

AFTER committing this: *No* popups from layer shell clients appear.
Example: hovering over waybar elements *never* generates any popups.

dwl.c
```diff
@@ -1606,7 +1611,7 @@ pointerfocus(Client *c, struct wlr_surface *surface, double sx, double sy,
 		focusclient(c, 0);
 
 	/* If surface is NULL, clear pointer focus */
-	if (!surface) {
+	if (!surface || exclusive_focus != surface) {
```
Sevz17, line 1614: @fauxmight can you revert this line and tell me if it works?

fauxmight: No noticeable change.

fauxmight:
Separate from the above issue: if I click on waybar (which normally *does* respond to mouse hover, but does *not* respond to mouse clicks), mouse *clicks* on XDG surfaces no longer register. Mouse *hover* still changes focus (and keyboard input) from one XDG client to another, but *clicks* do not. I can find no way to get mouse clicks' effects back to XDG clients.

Keyboard interaction is unaffected.

sevz17:
And with the latest changes?

fauxmight:
Now mouse clicks *never* register in any XDG shell client even if no layer shell client is started.

sevz17:
@fauxmight, I forgot ask you again, how does this work on the present state?

PR #264, opened 2022-06-28
Checks for overflows for client max width and height

### LINKED ISSUE #263: dwl crashes when starting Slack with multiple monitors
## Info
dwl's commit: 72e0a560d9836c5e8658003f548203bcd722e565
wlroots version: 0.15.1
## Description
This only seems to crash dwl when using two monitors, and only when opening slack on a particular of those monitors. When opened on the other, slack crashes, but dwl does not. I'm running slack on Electron 18 (which runs with native wayland support). This was introduced with 72e0a560d9836c5e8658003f548203bcd722e565, and is fixed by reverting to the previous commit.

I get the following error when in stderr when it crashes:
```
dwl: render/wlr_renderer.c:155: wlr_render_rect: Assertion `box->width > 0 && box->height > 0' failed.
```

BenJarg:
Checks whether `min.width + (2 * c->bw)` and `min.height + (2 * c->bw)` causes overflow, which can cause negative width and height and crash some clients. Closes #263.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -4,6 +4,7 @@
 #define _POSIX_C_SOURCE 200809L
 #include <getopt.h>
 #include <libinput.h>
+#include <limits.h>
 #include <linux/input-event-codes.h>
 #include <signal.h>
 #include <stdio.h>
@@ -386,9 +387,9 @@ applybounds(Client *c, struct wlr_box *bbox)
 	/* try to set size hints */
 	c->geom.width = MAX(min.width + (2 * c->bw), c->geom.width);
 	c->geom.height = MAX(min.height + (2 * c->bw), c->geom.height);
-	if (max.width > 0)
+	if (max.width > 0 && !(2 * c->bw > INT_MAX - max.width)) // Checks for overflow
 		c->geom.width = MIN(max.width + (2 * c->bw), c->geom.width);
-	if (max.height > 0)
+	if (max.height > 0 && !(2 * c->bw > INT_MAX - max.height)) // Checks for overflow
 		c->geom.height = MIN(max.height + (2 * c->bw), c->geom.height);
 
 	if (c->geom.x >= bbox->x + bbox->width)


dwl.c
```diff
@@ -387,9 +388,15 @@ applybounds(Client *c, struct wlr_box *bbox)
 	c->geom.width = MAX(min.width + (2 * c->bw), c->geom.width);
 	c->geom.height = MAX(min.height + (2 * c->bw), c->geom.height);
 	if (max.width > 0)
-		c->geom.width = MIN(max.width + (2 * c->bw), c->geom.width);
+		if ((2 * c->bw) > INT_MAX - max.width) // Checks for overflow
+			c->geom.width = c->geom.width;
+		else			
+			c->geom.width = MIN(max.width + (2 * c->bw), c->geom.width);
```
sevz17, line 394:
```suggestion
	if (max.width > 0 && !(2 * c->bw > INT_MAX - max.width)) // Checks for overflow
		c->geom.width = MIN(max.width + (2 * c->bw), c->geom.width);
```

dwl.c
```diff
@@ -387,9 +388,15 @@ applybounds(Client *c, struct wlr_box *bbox)
...
 	if (max.height > 0)
-		c->geom.height = MIN(max.height + (2 * c->bw), c->geom.height);
+		if ((2 * c->bw) > INT_MAX - max.height) // Checks for overflow
+			c->geom.height = c->geom.height;
+		else			
+			c->geom.height = MIN(max.height + (2 * c->bw), c->geom.height);
```
sevz17, line 399:
```suggestion
	if (max.height > 0 && !(2 * c->bw > INT_MAX - max.height)) // Checks for overflow		
		c->geom.height = MIN(max.height + (2 * c->bw), c->geom.height);
```

sevz17 (changes requested):

sevz17:
Thanks!

PR #267, opened 2022-07-12
Force line-buffered stdout if stdout is not a tty

dimkr:
The default buffering mode (chunk/line/none) of stdout is determined by stat(STDOUT_FILENO) in libc: if dwl is started under a tty, printf() output is written to STDOUT_FILENO immediately. Under `dwl > /tmp/log`, stdout is chunk-buffered.

This becomes a problem when redirection is used together with `-s`: after the dup2(), STDOUT_FILENO is connected to the stdin of a status bar (which needs to read dwl status updates immediately), while printf() output may be chunk-buffered (because STDOUT_FILENO wasn't a tty).


sevz17:
I'm not sure about this, [fflush()](https://github.com/djpohly/dwl/blob/main/dwl.c#L1673) does this job. and the time to complete the loop is negligible

sevz17:
@dimkr, I changed my mind, please remove the call to `fflush()`

dimkr:
@sevz17 What made you change your mind?

I'm currently debugging some voodoo related to this PR but I kinda ruled out the possibility that dwl is the issue. I have a weird setup: `dwl -s` runs a tool that sends every line it reads from stdin to a datagram Unix socket, so it's non-blocking for dwl. A panel started by this tool (with `< /dev/null`, but this doesn't matter) listens on the Unix socket path and receives the data. For some reason, sometimes the active window title in the panel "lags" behind: I switch from window x to window y, and the panel displays the title of x. When I switch from y to z, it displays the title of y. This happens when I switch between tags, too. Anyway, `strace` says the panel receives the data, and the bug should be somewhere after dwl.

sevz17:
> @sevz17 What made you change your mind?

It doesn't hurt to print lines one by one.

sevz17:
Thanks!

PR #268, opened 2022-07-13
Highlighting dependencies due to their invisibility.

David-Valters:
It's hard to see which dependencies are needed.


sevz17:
Thanks!

PR #276, opened 2022-07-23
use level 0 keysyms

sevz17:
this allow do not have to change the keysym when WLR_MODIFIER_SHIFT is active

NOTE: currently this patch makes ctrl+alt+F# not work, as well as ctrl+alt+backspace

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -88,11 +88,11 @@ static const double accel_speed = 0.0;
 
 /* If you want to use the windows key change this to WLR_MODIFIER_LOGO */
 #define MODKEY WLR_MODIFIER_ALT
-#define TAGKEYS(KEY,SKEY,TAG) \
+#define TAGKEYS(KEY,TAG) \
 	{ MODKEY,                    KEY,            view,            {.ui = 1 << TAG} }, \
 	{ MODKEY|WLR_MODIFIER_CTRL,  KEY,            toggleview,      {.ui = 1 << TAG} }, \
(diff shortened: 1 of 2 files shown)


fauxmight:
The loss of ctrl+alt+F# and ctrl+alt+backspace seems greater than the benefit of usable level-0 keysyms.

Perhaps this is more appropriate as a patch.
Or perhaps merging [sevz17/keycodes](https://github.com/sevz17/dwl/tree/keycodes) would be preferable.

sevz17:
> The loss of ctrl+alt+F# and ctrl+alt+backspace seems greater than the benefit of usable level-0 keysyms.

This is mostly to match dwm behavior but you're right make it usable would require more work (I'm not sure if it's suitable change to XKB_KEY_F# and XKB_KEY_BackSpace)

> Perhaps this is more appropriate as a patch. Or perhaps merging [sevz17/keycodes](https://github.com/sevz17/dwl/tree/keycodes) would be preferable.

I have no plans to merge keycodes, (get dwm's folks to use keycodes and I will merge it)

dimkr:
https://github.com/djpohly/dwl/compare/main...dimkr:dwl:en-keycodes.patch makes key bindings work when using a non-English layout, and it should be a backward-compatible change.

@sevz17 Would you consider this patch? (Not a replacement for this PR, of course)

sevz17:
Sorry but no. parity with dwm is one of our goals (that's why I won't merge keycodes) and dwm doesn't (and I think it can't) use a specific keymap to handle keybindings

pm4rcin:
I think that switching TTYs is important feature that should work no matter what. It's not used very often but when something goes wrong it's possible to kill bad processes.

PR #277, opened 2022-07-24
Fix segfault if parent->mon is unset

dimkr:
I don't know why, but since 90a12c9, mtpaint triggers a crash: c->mon is NULL after mapnotify(). This application has two windows, a main window and a small window with drawing tools.

This application hardcodes GDK_BACKEND=x11, so it's built with a patch (https://github.com/puppylinux-woof-CE/woof-CE/blob/testing/woof-code/rootfs-petbuilds/mtpaint/wayland.patch).

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1437,7 +1437,8 @@ mapnotify(struct wl_listener *listener, void *data)
 		/* Set the same monitor and tags than its parent */
 		c->isfloating = 1;
 		wlr_scene_node_reparent(c->scene, layers[LyrFloat]);
-		setmon(c, p->mon, p->tags);
+		/* TODO recheck if !p->mon is possible with wlroots 0.16.0 */
+		setmon(c, p->mon ? p->mon : selmon, p->tags);
 	} else {
 		applyrules(c);
 	}


sevz17:
Please add a comment `TODO: recheck with wlroots-0.16`

sevz17:
Actually according to the xdg-protocol if a surface set its parent to a unmapped surface is equal to set no parent (which I guess is happening here, that's why parent->mon is null). I think in wlroots-0.16 it is fixed, thus in theory that line even must not be reached.

Anyway, can you check if dwl segfaults in [sevz17/v0.4.x](https://github.com/sevz17/dwl/tree/v0.4.x) please?

sevz17:
Thanks!

dimkr:
> Anyway, can you check if dwl segfaults in [sevz17/v0.4.x](https://github.com/sevz17/dwl/tree/v0.4.x) please?

Will do, at some point in the future. I'm currently tracking both Debian Bookworm and Sid, so I'll get a chance to address this TODO when wlroots 0.16.0 lands in Debian.

PR #283, opened 2022-08-04
Respect rootcolor[]

dimkr:


sevz17:
I don't feel this belongs to mainline. probably to get this in dwl we want to create an mr to wlroots to allow set the root color.

fauxmight:
If rootcolor is currently meaningless, should it not be at the very least commented upon in `config.def.h` or, better yet, removed from `config.def.h`?

PR #284, opened 2022-08-05
2200: let's use this wisely

### LINKED ISSUE #278: increase the limit of SLOC
We are running out of lines (at b04c73be3de62d2739b5bb85e40f0c9af1122903 tokei reports exactly 2000 SLOC)

So the question is: does anyone really need it?

djpohly:
Hopefully fixes #278.


sevz17:
I think this is sufficient. LGTM

PR #285, opened 2022-08-09
Implement wlr_relative_pointer_manager_v1

BenJarg:
This makes games such as Minecraft function properly. As is, moving the mouse causes some spiraling behavior which makes the game unplayable. This implementation is similar to that of Sway's.


PalanixYT:
This already exists as a patch.
https://github.com/djpohly/dwl/compare/main...PalanixYT:point.patch

PR #287, opened 2022-08-11
rework outputmgrapplyortest()

sevz17:
first disable requested monitors, then enable and/or change mode, x and y, etc.
This is mostly what sway does

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1621,7 +1621,23 @@ outputmgrapplyortest(struct wlr_output_configuration_v1 *config, int test)
 			ok &= wlr_output_test(wlr_output);
 			wlr_output_rollback(wlr_output);
 		} else {
-			ok &= wlr_output_commit(wlr_output);
+			int output_ok = 1;
+			/* If it's a custom mode to avoid an assertion failed in wlr_output_commit()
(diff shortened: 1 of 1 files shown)


sevz17:
I tested and with this I am able to change to a a non-custom mode, when i tried to change to a custom mode dwl segfaults in wlr_output_commit()

sevz17:
> With this I am able to change to a a non-custom mode, if I try to change to a custom mode dwl segfaults in wlr_output_commit()

seems like https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3641 could fix the segfault setting a custom mode.

dimkr:
> This is mostly what sway does

How about integrating https://github.com/djpohly/dwl/compare/main...dimkr:dwl:fallback.patch?

sevz17:
> > This is mostly what sway does
> 
> How about integrating https://github.com/djpohly/dwl/compare/main...dimkr:dwl:fallback.patch?

I see the use case, however I don't think it should be on the mainline, but I am not strongly opposed to merging it.

Please open a pr, let @djpohly decide if we merge it.

PR #291, opened 2022-08-14
Fix disabled mons v2

### LINKED ISSUE #289: disabling monitor outputs don't work that well
## Info
dwl's commit: 0.3.1-74-g7a343b9-dirty
wlroots version: 0.15.1-6
## Description
Hello I have 2 monitor LVDS-1 and HDMI-1
I've noticed that when I disable the LVDS-1 monitor with "wdisplays" or typing `wlr-randr --output LVDS-1 --off` it disables the monitor but the clients don't move to the active HDMI-1 monitor they stay in the disabled LVDS-1 monitor.
The mouse pointer doesn't stay contained in the HDMI-1 monitor it goes through that disabled monitor.

### Opinion
Maybe the culprit is in the `cleanupmon()` function or the `closemon()` function.
I don't know how to use gdb to debug dwl so I might be wrong.

sevz17:
Fixes #289.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1199,9 +1199,12 @@ focusclient(Client *c, int lift)
 void
 focusmon(const Arg *arg)
 {
-	do
+	Monitor *m = selmon;
+	do {
 		selmon = dirtomon(arg->i);
-	while (!selmon->wlr_output->enabled);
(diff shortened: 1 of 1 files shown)


sevz17:
Closed in favor of #290

Issue #292, opened 2022-08-15
XWayland clients are rendered incorrectly when on fullscreen


pm4rcin:
## Info
dwl's commit: https://github.com/djpohly/dwl/commit/7a343b98cf37fb293313dada734c0d433b27fba4
wlroots version: 0.15.1
## Description
The wallpaper is visible under XWayland clients on edges.
![1660581111-wayshot](https://user-images.githubusercontent.com/37148802/184676538-47c5626c-b9ef-43cd-8c25-7930bcf0ebfb.png)

PR #293, opened 2022-08-17
don't respect size hints for fullscreen clients

sevz17:
Fixes: #292


pm4rcin:
Patch works as expected. Thanks! :)

PR #295, opened 2022-08-19
prevents natural_scrolling configuring usb mouse

DarioTD:
Now natural_scrolling only configures touch supported devices like a laptop touchpad.
Leaving the usb mouse that have the capability unaffected.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1004,10 +1004,9 @@ createpointer(struct wlr_input_device *device)
 			libinput_device_config_tap_set_enabled(libinput_device, tap_to_click);
 			libinput_device_config_tap_set_drag_enabled(libinput_device, tap_and_drag);
 			libinput_device_config_tap_set_drag_lock_enabled(libinput_device, drag_lock);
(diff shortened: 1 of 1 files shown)


fauxmight:
I use natural scrolling with a conventional USB mouse.
Making it possible to set natural_scrolling PER DEVICE sounds appropriate.
Forcing it impossible to use natural scrolling with a conventional mouse is a regression.

Perhaps this is more appropriate as a patch than in mainline.

DarioTD:
> I use natural scrolling with a conventional USB mouse. Making it possible to set natural_scrolling PER DEVICE sounds appropriate. Forcing it impossible to use natural scrolling with a conventional mouse is a regression.
> 
> Perhaps this is more appropriate as a patch than in mainline.

I agree but I don't know how to make it per input device class.
I will close this and maybe create a struct in dwl.c like MonitorRule struct and have an array of input device classes and their configurations.
When I get familiar with programming and wlroots of course.

PR #311, opened 2022-09-28
use sigaction(2) for signal handling

sevz17:
which means replace signal(2) with sigaction(2)

TODO: find a better description

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -1885,6 +1885,8 @@ run(char *startup_cmd)
 {
 	/* Add a Unix socket to the Wayland display. */
 	const char *socket = wl_display_add_socket_auto(dpy);
+	struct sigaction sa = {.sa_flags = SA_RESTART, .sa_handler = SIG_IGN};
+	sigemptyset(&sa.sa_mask);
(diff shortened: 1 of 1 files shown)


djpohly:
Rationale for the original: dwm does it.

sevz17:
> Rationale for the original: dwm does it. (Just as a note, not that I disagree with the change. 

Yes, they are discussing it in their [mailing list](https://lists.suckless.org/hackers/2207/18405.html); after reading [signal(2)](https://man7.org/linux/man-pages/man2/signal.2.html) I think we should switch to [sigaction(2)](https://man7.org/linux/man-pages/man2/sigaction.2.html) (primarily because we ignore SIGPIPE) but nobody has reported problems because that, so we could wait to see what dwm devs do.

sevz17:
Alternatively we can use the wayland helper: `wl_event_loop_add_signal()`

sevz17:
@djpohly, dwm started using sigaction() recently, I will wait two weeks in case they revert the commit or if you want to say something, if not I will merge this on Feb 13th.

PR #320, opened 2022-10-08
Allow moveresize() while the left mouse button is pressed (fixes #319)

### LINKED ISSUE #319: b8bc54 changes the behavior of moveresize()
I use a dwl patch that adds request_move and request_maximize handlers, so GTK+ title bar actions work. request_move is handled by calling moveresize().

May patch is broken since this commit, because moveresize() doesn't do anything when a mouse button is pressed.

Before CurPressed was introduced, moveresize() was allowed when the left mouse button is pressed. Is there any reason not to restore the previous behavior?

```
 moveresize(const Arg *arg)
 {
-	if (cursor_mode != CurNormal)
+	if (cursor_mode != CurNormal && cursor_mode != CurPressed)
```

dimkr:


sevz17:
Thanks!

PR #322, opened 2022-10-15
Add basic touch screen support

Unprex:
Hi, I don't know if this should be a patch, but it can serve as a base for other touch functionality.
Right now it redirects touch events to the pointer, emulating a mouse.

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -86,6 +86,11 @@ LIBINPUT_CONFIG_ACCEL_PROFILE_ADAPTIVE
 static const enum libinput_config_accel_profile accel_profile = LIBINPUT_CONFIG_ACCEL_PROFILE_ADAPTIVE;
 static const double accel_speed = 0.0;
 
+/* Touch screen */
+static const bool touch_enabled = true;
+static const int32_t touch_finger = 0; /* multi-touch finger id */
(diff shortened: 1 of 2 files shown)


fauxmight:
@Unprex: Yes, this should definitely be a patch because
1) IT'S AWESOME! Thank you! I would never have gotten around to this.
2) It won't be mainlined because dwm doesn't handle touch (because xorg handles touch *for* dwm).

Unprex:
I've added the patch [touchscreen](https://github.com/djpohly/dwl/wiki/touchscreen).
I'll update it as I go to add more functionality. (Or should I keep it simple and do multiple patches?)

fauxmight:
Patches in dwl are normally used by those who like the feature you have added (as opposed to discrete small changes in a queue waiting to be merged into a project's mainline). It would make sense to just keep updating your branch being sure to regularly merge changes from djpoly/main and update any significant progress on the patches page you've just made.

I wouldn't remove the "touch_enabled" config. (You mentioned "will probably be removed" on the patch page). Leaving it in place means that if someone has a custom build that pulls in your patch, they can still easily temporarily disable touch by editing just their own config.h and recompiling.

Thank you, once again, for this very nice feature! If you don't do it first, I will look at adding multi-touch scrolling eventually.

PR #323, opened 2022-10-15
stop trying resize clients during commitnotify()

sevz17:
instead resize them in configurex11()


sevz17:
partially merged in a996191

PR #329, opened 2022-11-03
check null popup->parent

shua:
managed to SEGFAULT the server by trying to create a popup without setting a parent first.

Not sure if this is dwl or wlroots issue, so also opened a ticket upstream: https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/3811


sevz17:
Thanks for the patch!, and sorry for the long delay in merging it.

Issue #327, opened 2022-11-01
Unset fullscreen when new client is spawned


dm1tz:
Current behaviour: spawning new client along with fullscreen window makes this window remain behind newly created one.

https://user-images.githubusercontent.com/82608185/199267881-8ba4a73f-7ce5-418c-8bce-b455a1ab5ef5.mp4

Expected behaviour: when the new window is spawned unset the fullscreen client back to tiled.

https://user-images.githubusercontent.com/82608185/199268363-7d763d9e-3dde-4270-8225-1e74e075518c.mp4

sevz17:
Makes sense, we could also create a dedicated scene tree for fullscreen clients, this way when creating clients they will appear below the fullscreen clients. Or both

PR #334, opened 2022-11-11
merge wlroots-next

sevz17:
Closes: #307

PR #335, opened 2022-11-12
unset fullscreen when a new client is created.

sevz17:
Requires: #334
Fixes: #327


sevz17:
cherry-picked first commit

PR #338, opened 2022-11-23
Fix comparison between signed and unsigned int

krypciak:
When c->bw (border width) is set to 0, the right side of the MAX macro gets turned into an unsigned integer and that results in -1 being the outcome.
This causes big issues for xwayland clients, for example, not tiling.


sevz17:
Thanks!

PR #342, opened 2022-12-02
Replace alacritty with foot in dwl(1) manpage

jbeich:


jbeich:
[Applied downstream](https://github.com/freebsd/freebsd-ports/commit/7c9c3e9ab546), mainly for binary package with default config.

sevz17:
Thanks!

PR #344, opened 2022-12-03
Fix idle inhibitor crash

sevz17:
@BenJarg can you test this?


BenJarg:
Looks like that takes care of it

Issue #306, opened 2022-09-16
Flickering when modifying main area's ratio


chcore:
## Info
dwl's commit: ba7dcb2dea2d5dc46b73b6b9194daab94a325a05 (master at time of writing, but also happens with the v0.3.1 release)
wlroots version: 0.15.1
## Description

There is noticeable flickering when resizing the ratio of the main area. I'm using the default config, with no patches or modifications. Not shown in the video is the flicker overflowing onto the next display on the right, when increasing the ratio.

https://user-images.githubusercontent.com/83099052/190829920-942f20ae-b38f-4e21-92ce-4155a23ab40a.mp4

Note that the above video happens to show dwl running inside another compositor, but the exact same thing happens when running it standalone.

Thanks

PalanixYT:
That behaviour is known and a fix is already almost finished.

pm4rcin:
@PalanixYT any progress or a patch to test?

PalanixYT:
Yeah you can look at v2flick branch, there is the fix. It was working great until firefox started acting up, refusing to update unless given focus (this requires you to have 2 monitors) to reproduce you spawn firefox on monitor 1 and then spawn another client, like foot, on monitor 2. then move the extra client to monitor 1. firefox will still be fine, display will update. Then, repeat this step and move another client, the display will freeze.
tldr; no idea what I'm doing, @sevz17 now does it

PalanixYT:
Turns out it was just a problem with wlroots because merging it with wlroots-next works.

sevz17:
Fixed in 7eaa01ac1f074511ae1013326172d51c6fdf8866

chcore:
This is not fixed if your `repeat_rate` and `repeat_delay` are fast enough. I use these:

`static const int repeat_rate = 50;`
`static const int repeat_delay = 300;`

And am seeing the exact same issue. For comparison, Sway does not have this problem when configured with the same repeat and delay settings.

sevz17:
> This is not fixed if your `repeat_rate` and `repeat_delay` are fast enough. I use these and am still seeing the exact same issue on master:
> 
> `static const int repeat_rate = 50;` `static const int repeat_delay = 300;`
> 
> For comparison, Sway does not have this problem when configured with the same repeat and delay settings.

Actually, this will not be 100% fixed until scene-graph implements atomic commits.

chcore:
@sevz17 is that feature tracked anywhere?

Issue #349, opened 2022-12-05
Freeze when spawing floating window in tiling layout


maxagabb:
## Info
dwl version: HEAD 16a49e99557563252b9f91db767b431e2238f587
wlroots version: 0.16
## Description
Hello! dwl freezes for me when spawning a floating window inside tiling layout. Seems to freeze longer if there are multiple tiled windows. Also freezes if I try to resize it. 

For example launching alacritty like so:
`setsid -f alacritty --class float`
```
static const Rule rules[] = {
	/* app_id     title       tags mask     isfloating   monitor */
	{ "float",    NULL,       0,            1,           -1 },
	{ "firefox",  NULL,       1 << 8,       0,           -1 },
};
```
 **EDIT:** Issue seems to be specific to alacritty

sevz17:
Probably it was introduced by 017bb7d7521f68d37bfe656c10f45edbcc92dd61, can you revert it and test?

maxagabb:
Yup, that fixes the issue completely

sevz17:
It wasn't a freeze for me, the only thing I noticed is that I couldn't move the mouse, key presses seemed fine.

maxagabb:
Just tried this again with `exec dwl` (no startup script). Also I have two monitors.

Issue did not happen with foot terminal, but I saw a freeze when resizing alacritty after launching it as floating.

I'm using alacritty-sixel-git from the AUR so that could be an issue.

maxagabb:
Tried this again with alacritty from normal repo and dwl without startup script.

Issue happens when there are multiple tiled alacritty windows. Keyboard input seems to work (I can close windows to get back the mouse)

maxagabb:
Now I feel this not a dwl bug since foot works fine. It's probably some issue in alacritty-wlroots that can be replicated with https://github.com/djpohly/dwl/commit/017bb7d7521f68d37bfe656c10f45edbcc92dd61

Should I close this?

sevz17:
Testing again I discovered that the cursor is perfectly functional, for some reason alacritty (or dwl?, time ago I did a commit related to it) hide it, foot doesn't.

Job79:
I am also experiencing "freezes" after the latest updates. It is pretty simple to reproduce:
1. Open 2 chromium or alacritty windows on a tag (doesn't seem to work when using foot)
2. Switch to another tag and open some other applications
3. Go bag to the tag with chromium/alacritty

[...]

PR #350, opened 2022-12-06
fix flickering

sevz17:
Before merging, make sure #349 is fixed.

Fixes: #306

### DIFF
--- a/client.h
+++ b/client.h
@@ -241,6 +241,21 @@ client_is_mapped(Client *c)
 }
 
 static inline int
+client_is_rendered_on_mon(Client *c, Monitor *m)
+{
+	/* This is needed for when you don't want to check formal assignment,
+	 * but rather actual displaying of the pixels.
+	 * Usually VISIBLEON suffices and is also faster. */
+	struct wlr_surface_output *s;
+	if (!c->scene->node.enabled)
+		return 0;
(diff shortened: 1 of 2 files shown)


sevz17:
@Job79, @PalanixYT, can you test this?

Job79:
Can no longer reproduce the "freeze" issue from #349 when applying this pull request

Job79:
Have been using this patch for 2 days now, never experienced any freezes. But I did see some flickering when resizing / opening alacritty. There is no flickering when using foot / chromium, so this is probably an alactitty problem?

sevz17:
> Have been using this patch for 2 days now, never experienced any freezes. But I did see some flickering when resizing / opening alacritty. There is no flickering when using foot / chromium, so this is probably an alactitty problem?

Not really, this is hackish, we should listen to the `commit` event, not `ack_configure`, but some clients do not call wl_surface.commit when we resize them with the same size (and this is ok, they are not obligated to do it), wl_surface.commit must be called after submitting a new buffer (e.g because it changed its size, its content, etc), `ack_configure` is just to let us know that they received the configure (new size) event (at this moment they aren't obligated to submit a new buffer)

sevz17:
Superseded by #357 and #363

PR #357, opened 2022-12-12
fix flickering v2

### LINKED ISSUE #306: Flickering when modifying main area's ratio

sevz17:
Alternative to #350.

Fixes #306

### DIFF
--- a/client.h
+++ b/client.h
@@ -241,6 +241,21 @@ client_is_mapped(Client *c)
 }
 
 static inline int
+client_is_rendered_on_mon(Client *c, Monitor *m)
+{
+	/* This is needed for when you don't want to check formal assignment,
+	 * but rather actual displaying of the pixels.
+	 * Usually VISIBLEON suffices and is also faster. */
+	struct wlr_surface_output *s;
+	if (!c->scene->node.enabled)
+		return 0;
+	wl_list_for_each(s, &client_surface(c)->current_outputs, link)
+		if (s->output == m->wlr_output)
+			return 1;
+	return 0;
+}
+
+static inline int
 client_is_unmanaged(Client *c)
 {
 #ifdef XWAYLAND
--- a/dwl.c
+++ b/dwl.c
@@ -109,6 +109,7 @@ typedef struct {
 		struct wlr_xdg_surface *xdg;
 		struct wlr_xwayland_surface *xwayland;
 	} surface;
+	struct wl_listener ack_configure;
 	struct wl_listener commit;
 	struct wl_listener map;
 	struct wl_listener maximize;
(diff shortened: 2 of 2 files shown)


sevz17:
@Job79, this fixes the flickering with Alacritty, seems like this doesn't cause freezes, I will test during the next days, can you test too?

Job79:
Just applied the diff to my local setup, all seems to work fine so far. Will report back when I experience any problems during the next days. Thanks for working on these patches, been daily driven dwl for 6 months now and am really liking it.

Job79:
Can confirm the issue with Xwayland applications when using these changes. When using phpstorm(Xwayland application) the buttons in the menu bar no longer work. They works fine when using the previous "fix flickering" PR.

maxagabb:
This PR feels really nice! Flickering in alacritty still feels fully resolved with https://github.com/djpohly/dwl/commit/a9b1fbf652ad4657d4c7013ee3238bedb691408c reverted. Xwayland menus also work again.

Alternatively re-enabling scene node inside #ifdef also fixes Xwayland menus when I test:
```
#ifdef XWAYLAND
	/* Handle unmanaged clients first so we can return prior create borders */
  wlr_scene_node_set_enabled(&c->scene->node, 1);
```

sevz17:
> This PR feels really nice! Flickering in alacritty still feels fully resolved with [a9b1fbf](https://github.com/djpohly/dwl/commit/a9b1fbf652ad4657d4c7013ee3238bedb691408c) reverted. Xwayland menus also work again.

That commit is meant to fix flickering when mapping clients.

> Alternatively re-enabling scene node inside #ifdef also fixes Xwayland menus when I test:
> 
> ```
> #ifdef XWAYLAND
> 	/* Handle unmanaged clients first so we can return prior create borders */
>   wlr_scene_node_set_enabled(&c->scene->node, 1);
> ```

I pushed an alternative solution to only disable the node for XDG clients.

sevz17:
Also, I found a freeze when running a nested dwl and it is stopped (e.g because you are debugging it ) the host dwl will freeze when switch to another tag and then switch back.

I'll try a v3 of this, the problem above probably never will happen to normal users but I constantly have to debug dwl, and honestly I don't want to merge this in this state, even if it fixes the flickering.

maxagabb:
Makes a lot of sense, also thanks for all the improvements! I'm having great time daily driving dwl.

sevz17:
Superseded by #363

PR #362, opened 2022-12-15
Check width and height are not negative in client_set_bounds()

BenJarg:
`wlr_xdg_toplevel_set_bounds()` asserts width and height >= 0, so if you attempt to resize a native wayland window too far left or up ending up with a negative height or width, dwl crashes with a `SIGABRT`.


sevz17:
Thanks!

PR #363, opened 2022-12-16
fix flickering v3

sevz17:
This mitigate the issue mentioned [here](https://github.com/djpohly/dwl/pull/357#issuecomment-1352518941) and also is shorter than #357

### DIFF
--- a/client.h
+++ b/client.h
@@ -241,6 +241,21 @@ client_is_mapped(Client *c)
 }
 
 static inline int
+client_is_rendered_on_mon(Client *c, Monitor *m)
+{
+	/* This is needed for when you don't want to check formal assignment,
+	 * but rather actual displaying of the pixels.
+	 * Usually VISIBLEON suffices and is also faster. */
+	struct wlr_surface_output *s;
+	if (!c->scene->node.enabled)
+		return 0;
+	wl_list_for_each(s, &client_surface(c)->current_outputs, link)
+		if (s->output == m->wlr_output)
+			return 1;
+	return 0;
+}
+
+static inline int
 client_is_unmanaged(Client *c)
 {
 #ifdef XWAYLAND
--- a/client.h
+++ b/client.h
@@ -256,6 +256,32 @@ client_is_rendered_on_mon(Client *c, Monitor *m)
 }
 
 static inline int
+client_is_stopped(Client *c)
+{
+	int pid;
+	siginfo_t in = {0};
+#ifdef XWAYLAND
+	if (client_is_x11(c))
+		return 0;
+#endif
+
+	wl_client_get_credentials(c->surface.xdg->client->client, &pid, NULL, NULL);
+	if (waitid(P_PID, pid, &in, WNOHANG|WCONTINUED|WSTOPPED|WNOWAIT) < 0) {
+		/* This process is not our child process, while is very unluckely that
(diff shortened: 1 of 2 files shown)


BenJarg:
I'm still seeing an issue like the one you mentioned in #357. For example, starting a client, putting it in another tag by itself, then run `kill -STOP PID` on the process, when I try to switch to the tag with the stopped window, dwl doesn't update until switching to a different tag.

sevz17:
> I'm still seeing an issue like the one you mentioned in #357. For example, starting a client, putting it in another tag by itself, then run `kill -STOP PID` on the process, when I try to switch to the tag with the stopped window, dwl doesn't update until switching to a different tag.

Yep, that's why I said "mitigate", there is no easy way to do not skip frames if the client is stopped.

sevz17:
I think I won't merge this or #357 for 0.4

sevz17:
@BenJarg, @Job79, @maxagabb, can you try this?, I was able to deal with stopped clients.

maxagabb:
On this PR I was able to replicate the stopped client issue (dwl not updating) with BenJarg's steps using ` kill -STOP PID `

On main branch the behavior is different. dwl updates but the stopped client is not sized correctly;

sevz17:
> On this PR I was able to replicate the stopped client issue (dwl not updating) with BenJarg's steps using `kill -STOP PID`

Very odd, FWIW it works fine to me.

> On main branch the behavior is different. dwl updates but the stopped client is not sized correctly

That's the expected (and only possible) behavior, the client can do nothing.

maxagabb:
Yup could be an issue with how I'm testing. I tried removing all files in .gitignore and compiling again.

I use an alacritty window to send STOP to foot after sending foot to another tag.

Issue does not happen if I go the foot tag once before sending STOP from alacritty.

sevz17:
> Yup could be an issue with how I'm testing. I tried removing all files in .gitignore and compiling again.
> 
> I use an alacritty window to send STOP to foot after sending foot to another tag.
> 
> Issue does not happen if I go the foot tag once before sending STOP from alacritty.

It should be fixed now.

maxagabb:
Cool I confirmed this fixes the issue for me. 

I noticed there's flicker with foot terminals only when spawning them with bemenu (spawning and resizing).

When launched with keyboard shortcut they only flicker when spawned (no flicker on resize).

Job79:
Am testing this patch on my local dwl, seems to work fine so far. Never experienced issues with clients freezing when stopped, does this issue only occur when using `kill -STOP {pid}`?

maxagabb:
> Am testing this patch on my local dwl, seems to work fine so far. Never experienced issues with clients freezing when stopped, does this issue only occur when using `kill -STOP {pid}`?

Yep I believe that issue only came from `kill -STOP {pid}`, no freezes otherwise

sevz17:
Yep, for processes where dwl is not their parent (bemenu makes double fork, which means that the created process is child of init) we assume that they are stopped (because we can't get their info), otherwise we could end up with a frozen display.

maxagabb:
I'm getting no flicker from alacritty windows now using the shortcut. 

The desktop background appears briefly when spawning them, but there's no flicker on second monitor (which happened before this PR). 

I'm seeing no issues at all when resizing.

BenJarg:
@sevz17 The stopped client issue looks resolved for me as well.

PR #373, opened 2023-01-15
Switch from Make build system to Meson

StratusFearMe21:
meson is a faster build system than Make and is easier to write. It also has more options that the user can set, for example, the user can add `-Dbuildtype=release` to the command line to make an optimized build.


.gitignore
```diff
@@ -4,3 +4,4 @@ dwl
 *-protocol.h
 .ccls-cache
 config.h
+build
```
eli-schwartz, line 7: https://mesonbuild.com/FAQ.html#how-do-i-ignore-the-build-directory-in-my-vcs

This is not necessary with current versions of Meson.

You can probably delete the .gitignore altogether if it removes the Makefile which is the only thing creating those files...

protocols/meson.build
```diff
@@ -0,0 +1,20 @@
...
+wayland_protos_dep = dependency('wayland-protocols')
+wl_protocol_dir = wayland_protos_dep.get_pkgconfig_variable('pkgdatadir')
+wayland_scanner_code = generator(
+	wayland_scanner,
+	output: '@BASENAME@-protocol.c',
+	arguments: ['private-code', '@INPUT@', '@OUTPUT@'])
+wayland_scanner_client = generator(
+	wayland_scanner,
+	output: '@BASENAME@-protocol.h',
+	arguments: ['server-header', '@INPUT@', '@OUTPUT@'])
```
eli-schwartz, line 11: You might also be interested in https://mesonbuild.com/Wayland-module.html

StratusFearMe21:
@eli-schwartz, we meet again! Thanks for such a quick review, I'm on it!

protocols/meson.build
StratusFearMe21: I'm not sure this is possible, the wlroots includes require that the protocol file is named `@BASENAME@-protocol.h`, and the module you mentioned seems to only support naming it as `@BASENAME@-server-protocol.h`

eli-schwartz: Hmm, I see.

That's probably a simple matter of a tiny source code modification, but OTOH that *is* more than a build system switch.

StratusFearMe21: Well, the source code that requires this is brought in from `/usr/include/`, so that kinda complicates things

eli-schwartz: Oh huh, so it is.

This is probably the fault of https://github.com/mesonbuild/meson/pull/9985 which always calls the header `{name}-{side}-protocol.h`, even though it looks like the wayland intention is for `{name}-protocol.h` for the server and `{name}-client-protocol.h` for the client? Maybe this needs to be fixed.

meson.build
```diff
@@ -0,0 +1,43 @@
...
+)
+
+subdir('protocols')
+
+cc = meson.get_compiler('c')
+libm = cc.find_library('m')
+xcb = dependency('xcb', required : get_option('xwayland'))
+xlibs = dependency('xcb-icccm', required : get_option('xwayland'))
+wayland_server_dep = dependency('wayland-server')
+wlroots_dep = dependency('wlroots')
```
dimkr, line 12: `version: ['>=0.16.0', '<0.17.0']` maybe?

meson.build
```diff
@@ -0,0 +1,43 @@
...
+xkbcommon_dep = dependency('xkbcommon')
+libinput_dep = dependency('libinput')
+
+c_args = [
+  '-DWLR_USE_UNSTABLE',
+  '-D_POSIX_C_SOURCE=200809L',
+  '-DVERSION="@0@"'.format(meson.project_version())
+]
+
+if xcb.found() and xlibs.found()
```
dimkr, line 22: I think `and get_option('xwayland')` is missing here - I build my dwl with Xwayland disabled although I have these optional dependencies installed.

StratusFearMe21: If the xwayland option is disabled and I put the `get_option` on the dependencies, meson will automatically resolve them as not found

eli-schwartz: Indeed, Meson's feature options are specifically designed to guarantee that you can single-source your information as much as possible. Pass one to a dependency lookup, and if it's disabled then the dependency lookup will log "Dependency foobar not found: feature **xwayland** disabled". No need to add if/else logic to sometimes check for them and sometimes not. No need to gate other checks on both "was the dependency found" and "was the dependency configured to be used".

eli-schwartz: In fact, `and get_option('xwayland')` will reduce to `and <feature_object>` which doesn't have a boolean state so that's just an error. You'd have to use `and <feature_object>.allowed()` to get a valid boolean check.

sevz17:
Thanks for working on dwl!, however I can't merge this.

> meson is a faster build system than Make and is easier to write.

Yes, I personally prefer meson over autotools, but dwl tries to mimic dwm hence the use of a makefile, and I don't see the suckless guys using meson. (apart that in this project the use of meson would actually cause more work, because users will need to run `meson setup build && meson compile -C build` instead of a simple `make`, and I will need to reconfigure every time I need to enable/disable Xwayland)

> It also has more options that the user can set, for example, the user can add `-Dbuildtype=release` to the command line to make an optimized build.

Users can either export the desired CFLAGS or pass it through arguments to make, e.g `export CFLAGS='-O3 -pipe -march=native -flto' && make` (subsequents call of make don't need to re-export CFLAGS) or `make CFLAGS='-O3 -pipe -march=native -flto'`

StratusFearMe21:
Absolutely understandable.

PR #374, opened 2023-01-15
Switch from Make build system to Meson (wlroots-next)

StratusFearMe21:
Just a version of my previous PR, #373  that adds support for building from the wlroots git


sevz17:
Same reason that the other one.

PR #378, opened 2023-01-24
Fix focus-follows-cursor (sloppyfocus) for GTK4 applications

lx7:
GTK4 applications (nautilus, font viewer etc.) show a funny behaviour where the window gets the focus before the cursor enters the window. This is caused by a window surface that extends beyond it's bounding box (as defined by x, y, width and height).

With this change, pointerfocus returns after wlr_seat_pointer_notify_clear_focus and before focusclient if the cursor is not within the client's bounding box.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -123,7 +123,7 @@ typedef struct {
 	struct wl_listener set_hints;
 #endif
 	unsigned int bw;
-	unsigned int tags;
+	uint32_t tags;
 	int isfloating, isurgent, isfullscreen;
 	uint32_t resize; /* configure serial of a pending resize */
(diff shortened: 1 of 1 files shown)


sevz17:
Hi, I'm sorry for the late reply.

This does not consider when a button is held, thus breaks for example the scrollbar in a browser (move it with the cursor, w/ and w/o this patch, specifically move the cursor outside the browser window)

pm4rcin:
I can't move the cursor to the edge of the screen with this patch. Also it doesn't fix the original issue you mentioned. Just try to move cursor slowly to GTK4 app.

PR #383, opened 2023-02-04
feat: use level-0 keysym for keybindings

federico-carbone:
This will solve #276 issue with VT switching


sevz17:
Apart that this is just a copy of mi pr, see https://github.com/djpohly/dwl/pull/276#issuecomment-1193259543

federico-carbone:
I don't understant the loss of ctrl+alt+F# and ctrl+alt+BackSpace, can you explain?

PR #386, opened 2023-02-17
fix spelling + remove unecersarry spaces

godalming123:
This fixes a spelling and grammar issue I noticed while browsing the code. It also removes several places where their should only be 1 space instead of 2.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -98,7 +98,7 @@ typedef struct Monitor Monitor;
 typedef struct {
 	/* Must keep these three elements in this order */
 	unsigned int type; /* XDGShell or X11* */
-	struct wlr_box geom;  /* layout-relative, includes border */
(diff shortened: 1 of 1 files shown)


sevz17:
Fix the indentation please. We use tabs instead of spaces.

pm4rcin:
@godalming123 if you fix small things like that could you also trim the trailing whitespaces [here](https://github.com/djpohly/dwl/blob/f8373ccf2581fa481602bbb7528ba8871f6d5c85/config.def.h#L69-L71)?
Btw. @sevz17 maybe it's a good idea to create .editorconfig for dwl?

godalming123:
I have replaced the spaces with tabs and also removed several places in the source coce where their are 2 spaces which should be 1

sevz17:
Thanks!

PR #399, opened 2023-03-12
implement virtual_pointer protocol

youbitchoc:
Not sure if I should submit this to the wiki instead since nobody seems to be complaining about wayvnc


sevz17:
Please add this on the wiki, as you said no one has complained about this missing protocol.

PR #402, opened 2023-03-14
focused borders on active client after lockscreen

bencollerson:
After locking the screen the active client no longer has focus indicated by focuscolor borders. This patch ensures the exclusive_focus variable is set so focuscolor borders are applied to the active client.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -2435,6 +2435,7 @@ unlocksession(struct wl_listener *listener, void *data)
 {
 	SessionLock *lock = wl_container_of(listener, lock, unlock);
 	destroylock(lock, 1);
+	arrangelayers(selmon);
 }
 
 void


gozenka:
I do not have this issue with swaylock.

bencollerson:
You get your focused window borders coloured correctly after swaylock is unlocked on an unpatched main branch dwl build? Because I did confirm that the bug I was seeing was occurring with unpatched dwl main and swaylock. Maybe something else is going on?

bencollerson:
Sorry I was incorrect. I wrap swaylock with a screen dimming program before swaylock is launched. I assume this program is somehow interfering with dwl applying focus correctly. This does not appear to be a real bug with dwl. I will close this pull request.

gozenka:
Still, for clarification:
I tried it unpatched, but I am currently on the wlroots-next branch.

bencollerson:
I am using my own modified version of https://git.sr.ht/~emersion/chayang that wraps around swaylock and dims the screen before locking it and only exits after the screen is unlocked. My version is here: https://github.com/bencollerson/chayang. I may have made a mistake in how I have implemented my changes to the screen dimmer. My patch to dwl still fixes the problem for me.

PR #403, opened 2023-03-23
remove tag labels from dwl

bencollerson:
Tag labels are not used in dwl. Only the number of tags is important. Tag labels should be defined for each tag in whatever status bar is used.

Also tag mask should be an unsigned int so that documented limit of 31 tags does not cause an integer overflow.


sevz17:
Merged in 9d68554c59a886b641d27a364884fb461af2d4f1

sevz17:
Thanks!

bencollerson:
Thankyou and happy Easter.

PR #421, opened 2023-05-24
Prevent using a wlr_layer_surface after destroying it

minego:
If the `createlayersurface` callback is hit when `selmon` is NULL then `wlr_layer_surface` is destroyed by calling `wlr_layer_surface_v1_destroy()`. This change simply adds a return at that point, to prevent a crash caused by dereferencing the destroyed variable.

The crash in question can be reproduced easily by unplugging or turning off all monitors while running `swaync`, waiting a few seconds and then plugging them back in. It is likely possible to cause this with things other than `swaync`, but that was the application that triggered it for me.


sevz17:
Thanks!

minego:
Thank you!

PR #425, opened 2023-05-28
Set XCURSOR_SIZE to fix cursor changing size

NikitaIvanovV:
When I hover a cursor over various clients, the cursor changes its size. For example, it's normal size on foot and becomes bigger in qutebrowser. I figured out that these clients use `XCURSOR_SIZE` env variable to determine appropriate cursor size and when it's unset, they fallback to using some default value. I presume this default is not the same on all clients, this is why cursor size changes. This PR adds a `setenv` call to make sure `XCURSOR_SIZE` is set to the proper value.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -2254,6 +2254,7 @@ setup(void)
 	 * images are available at all scale factors on the screen (necessary for
 	 * HiDPI support). Scaled cursors will be loaded with each output. */
 	cursor_mgr = wlr_xcursor_manager_create(NULL, 24);
+	setenv("XCURSOR_SIZE", "24", 1);
 
 	/*
 	 * wlr_cursor *only* displays an image on screen. It does not move around


bencollerson:
This is a good tip and is working for me by setting XCURSOR_SIZE in my dwl start up script. 

Not sure about this particular implementation though as there may be users already setting this variable in their environment outside of dwl. This patch would just overwrite any existing environment setting with no way to use an externally set value.

sevz17:
> This is a good tip and is working for me by setting XCURSOR_SIZE in my dwl start up script.
> 
> Not sure about this particular implementation though as there may be users already setting this variable in their environment outside of dwl. This patch would just overwrite any existing environment setting with no way to use an externally set value.

Actually setting XCURSOR_{THEME,SIZE} it the compositor's job, the reason dwl doesn't set it is because it doesn't let the user to customize the cursor

sevz17:
Thanks!

bencollerson:
Haha I stand corrected. ;)

PR #434, opened 2023-06-16
fix startup_cmd SIGCHLD handler

djpohly:
Looks like ignored signals aren't reset by `exec()` calls.  If the startup command needs to fork and wait on child processes (e.g. a session manager like s6-svscan), the `SIG_IGN` that is still in place will interfere with waiting.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -2327,6 +2327,9 @@ void
 spawn(const Arg *arg)
 {
 	if (fork() == 0) {
+		struct sigaction sa = {.sa_flags = SA_RESTART, .sa_handler = SIG_DFL};
+		sigemptyset(&sa.sa_mask);
+		sigaction(SIGCHLD, &sa, NULL);
 		dup2(STDERR_FILENO, STDOUT_FILENO);
 		setsid();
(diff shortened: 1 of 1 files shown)


sevz17:
> Looks like ignored signals aren't reset by `exec()` calls.

So we could do this for spawn() too, right?

djpohly:
Good call. If it makes the code simpler/easier to understand, I guess we could forgo the `SIG_IGN` altogether and just set a handler, which *would* then be reset by `exec`.

sevz17:
How many lines would it add?

I am a bit reluctant to add more lines, in `wlroots-next` there are already 2230 lines.

sevz17:
We could also entirely give up in the SLOC limit and try to make the code more readable/_correct_, but still trying to do things in the simplest way.

I am only concerned about SLOC because it's one of the principles of dwl, but I would not mind if we remove the limit.

djpohly:
Oh, I'll watch my SLOC footprint for sure!  I had a hunch the suggestion would actually reduce lines, and it turns out it did.  This PR now reduces the count from `cloc dwl.c` by 5.

sevz17:
Thank you!

PR #440, opened 2023-06-21
Fix the original push patch

Abanoub8:
This is the push patch ported to work with the latest commit of dwl (68a17f9).

use the new focustop(selmon) function instead of the old selclient() function which has been deprecated by commit 72adab6.

I am open to the possibility of creating a new patch to eliminate the push.c and push.h and instead add the functions directly to the dwl.c file.


sevz17:
@djpohly

djpohly:
Thanks!

PR #441, opened 2023-06-21
vertile

Abanoub8:
This is the vertile patch ported to work with the latest commit of dwl (68a17f9).

use the new resize() function which takes a struct wlr_box type instead of the old resize() function which takes many more parameters.


sevz17:
Thanks for updating the patches!

Unfortunately, we do not accepts patches in the main branch, however you can submit patches in the [wiki](https://github.com/djpohly/dwl/wiki/Patches#submitting-patches), since most of the patches looked more like updates, you may create PR's in the repos of the respective authors, if the author seems inactive you can change the patch URL to point to your branches.

PR #443, opened 2023-06-25
Refactor: separate drag icon from layers array and Lyr enum

djpohly:
If we treat the drag icon as distinct from other layers (it doesn't have contents that are interactive, focusable, etc.), then we can iterate over layers meaningfully with a simple for loop.

ΔSLOC: -8

(I tried to make sure I caught everywhere that we were assuming the ZWLR_LAYER_SHELL constants were the same as our enum, but I wouldn't mind having that double-checked for safety.)

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -74,7 +74,7 @@
 /* enums */
 enum { CurNormal, CurPressed, CurMove, CurResize }; /* cursor */
 enum { XDGShell, LayerShell, X11Managed, X11Unmanaged }; /* client types */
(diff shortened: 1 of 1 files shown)


fauxmight:
I cannot see any obvious misses.
I have merged the no-drag-icon-layer branch and am running it without issue.
Will report back here with any regressions noted.

sevz17:
Looks pretty good. Lets wait one or two days for @fauxmight to report regressions, In the meanwhile I'll be testing this too.

sevz17:
Thank!

PR #446, opened 2023-06-29
tagcount should have been a #define

bencollerson:
Tagcount should have been set up as a #define rather than a const so it can be used to in array declarations etc.


sevz17:
I'm sorry for the late reply.

I don't feel like breaking user configuration again, just for external patches.
I know, it's a trivial change, but still I don't think it should be made.

At most, I'm ok with adding a \#define and defining the variable with the macro.

sevz17:
Thanks!

neuromagus:
What I mean too. If u need variable in array length - this is only ```define```, ```const int tagcount``` need a delete. "salt is salty".

sevz17:
Ah yeah, I didn't say it but I'm gonna revert the second commit, but I'll be when doing another change to the config

PR #448, opened 2023-07-03
Unify signal handling under wl_event_loop

djpohly:
Merge our signal handlers into a single function and let Wayland deal with all the struct sigaction stuff.

ΔSLOC: -3

(I took the opportunity to update/clarify the comment about XWayland while I was in the neighborhood.)

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -260,6 +260,7 @@ static void focusmon(const Arg *arg);
 static void focusstack(const Arg *arg);
 static Client *focustop(Monitor *m);
 static void fullscreennotify(struct wl_listener *listener, void *data);
+static int handlesig(int signo, void *data);
 static void incnmaster(const Arg *arg);
 static void inputdevice(struct wl_listener *listener, void *data);
 static int keybinding(uint32_t mods, xkb_keysym_t sym);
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -260,6 +260,7 @@ static void focusmon(const Arg *arg);
 static void focusstack(const Arg *arg);
 static Client *focustop(Monitor *m);
 static void fullscreennotify(struct wl_listener *listener, void *data);
+static int handlesig(int unused, void *data);
```
sevz17, line 263:
```suggestion
static int handlesig(int signo, void *data);
```

dwl.c
```diff
@@ -2129,18 +2140,15 @@ setup(void)
...
-
 	/* The Wayland display is managed by libwayland. It handles accepting
 	 * clients from the Unix socket, manging Wayland globals, and so on. */
 	dpy = wl_display_create();
+	eventloop = wl_display_get_event_loop(dpy);
+	wl_event_loop_add_signal(eventloop, SIGCHLD, handlesig, NULL);
+	wl_event_loop_add_signal(eventloop, SIGINT, handlesig, NULL);
+	wl_event_loop_add_signal(eventloop, SIGTERM, handlesig, NULL);
+	/* Avoid terminating if nobody is reading the status output */
+	wl_event_loop_add_signal(eventloop, SIGPIPE, handlesig, NULL);
```
sevz17, line 2151: I am not sure if we need to call `wl_event_source_remove`

djpohly: Ah, I think you're right.  It doesn't look like anything in the Wayland code cleans this up.  I'll try to figure out if there's a way to do this "correctly" while keeping delta <= 0.

sevz17:
Thanks!

PR #449, opened 2023-07-03
Gamma control -> patch

### LINKED ISSUE #447: Move `wlr_gamma_control_manager_v1` to patch
Gamma control is a protocol that not everyone uses/needs, but it was added to dwl core because the SLOC impact was negligible.  Since this is no longer the case with wlroots 0.17 (it now costs us 24 lines), I propose moving the `wlr_gamma_control_manager_v1` implementation into a patch for wlroots-next.

djpohly:
This protocol doesn't need to be in core dwl, and it now costs more than the 2 SLOC that it used to, so it makes sense for it to be a community-maintained patch at this point.

Closes #447.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -24,7 +24,6 @@
 #include <wlr/types/wlr_linux_dmabuf_v1.h>
 #include <wlr/types/wlr_export_dmabuf_v1.h>
 #include <wlr/types/wlr_fractional_scale_v1.h>
-#include <wlr/types/wlr_gamma_control_v1.h>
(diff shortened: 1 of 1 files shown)


sevz17:
I was actually waiting for [wlroots!4192](https://gitlab.freedesktop.org/wlroots/wlroots/-/merge_requests/4192) to be merged.

djpohly:
Oh, nice! If that brings the impact back down to a couple of lines, then it's all good.

PR #451, opened 2023-07-03
Line saver: follow `local = wlr->data = obj` pattern

djpohly:
There were still a couple of places where we could use this handy multiple assignment to save a line.

ΔSLOC = -3


sevz17:
Thanks!

PR #452, opened 2023-07-03
Line saver: `LISTEN_STATIC` macro

djpohly:
This parallels the `LISTEN` macro for statically allocated listeners, and it allows us to remove almost all of the global wl_listener declarations.

This also fixes a bug with the axisnotify listener, which was declared with a compound literal.  At block scope, [these have automatic storage duration](https://en.cppreference.com/w/c/language/compound_literal#Explanation), so the listener was no longer valid after setup() returned.  (The option to declare it `static` explicitly was standardized in C23, if that ever gains suckless traction.)

ΔSLOC: -27


sevz17:
Thanks!

PR #458, opened 2023-07-13
Make drag_icon a persistent scene node

djpohly:
If there is no current drag icon, this node will be empty, but we now have `drag_icon != NULL` as an invariant.  This allows us to eliminate a conditional, since there's no harm in moving an empty node's coordinates around with the pointer.

ΔSLOC: -1

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -331,6 +331,7 @@ static struct wl_event_source *sighandler[4];
 static struct wlr_backend *backend;
 static struct wlr_scene *scene;
 static struct wlr_scene_tree *layers[NUM_LAYERS];
+static struct wlr_scene_tree *drag_icon;
 /* Map from ZWLR_LAYER_SHELL_* constants to Lyr* enum */
 static const int layermap[] = { LyrBg, LyrBottom, LyrTop, LyrOverlay };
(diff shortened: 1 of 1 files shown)


dwl.c
```diff
@@ -2337,8 +2338,7 @@ startdrag(struct wl_listener *listener, void *data)
 	if (!drag->icon)
 		return;
 
-	drag->icon->data = icon = wlr_scene_subsurface_tree_create(&scene->tree, drag->icon->surface);
-	wlr_scene_node_place_below(&icon->node, &layers[LyrBlock]->node);
+	drag->icon->data = icon = wlr_scene_subsurface_tree_create(drag_icon, drag->icon->surface);
```
sevz17, line 2341: `icon` is not needed anymore.

sevz17: And for a sightly better type safety we might want to change it to:
```suggestion
	drag->icon->data = &wlr_scene_subsurface_tree_create(drag_icon, drag->icon->surface)->node;
```

djpohly:
Now that I look at it, the `motionnotify` call there seems like it wouldn't do anything besides update the drag icon position, so I've deleted that as well.

New ΔSLOC: -3

PR #463, opened 2023-07-22
Remove some redundant function calls

djpohly:
Spotted a couple of things that wlroots already does for us, and a couple of places where we were called `printstatus()` after making no changes.

Net ΔSLOC: -5


dwl.c
```diff
@@ -1531,7 +1534,6 @@ void
 maplayersurfacenotify(struct wl_listener *listener, void *data)
 {
 	LayerSurface *l = wl_container_of(listener, l, map);
-	wlr_surface_send_enter(l->layer_surface->surface, l->mon->wlr_output);
```
sevz17, line 1534: As far I remember this is needed.

sevz17: Nevermind, it's not needed, but I found a funny thing, seems like the surfaces don't receive the wl_pointer.enter at all (with and without send wl_suface.enter ourselves)

dwl.c
```diff
@@ -1808,9 +1810,6 @@ outputmgrapplyortest(struct wlr_output_configuration_v1 *config, int test)
 	else
 		wlr_output_configuration_v1_send_failed(config);
 	wlr_output_configuration_v1_destroy(config);
-
-	/* TODO: use a wrapper function? */
-	updatemons(NULL, NULL);
```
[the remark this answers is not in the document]
djpohly: Weird.  Do you remember what kind of changes didn't apply, and are we sure it's not a wlroots bug?  These lines came from cb01ce9, in case that helps jog your memory.

(I don't have an external on me right now, so I had to test using the wl backend instead of drm.)

dwl.c
djpohly: I discovered that swaynag is a better test (guessing dtao and bemenu are binding to older protocol versions), and there I'm still seeing both surface-enter-output and pointer-enter-surface:

```
$ WAYLAND_DEBUG=1 swaynag -m hello |& grep -Fw -e enter -e leave
[ 631375.417] wl_surface@10.enter(wl_output@8)
[ 633471.854] wl_pointer@9.enter(1155, wl_surface@10, 1476.41406250, 33.74218750)
[ 635047.065] wl_pointer@9.leave(1156, wl_surface@10)
[ 636230.590] wl_pointer@9.enter(1159, wl_surface@10, 1490.90625000, 32.12109375)
```

For the record, bemenu showed keyboard-enter with a similar test.

sevz17: Oh, I meant, it does not receive wl_pointer.enter until the cursor is moved, to reproduce:
1. run `WAYLAND_DEBUG=1 fnott`
2. move the cursor where the notification will be placed
3. send a notify

sevz17:
Merged "No need to send surface.leave/enter events" and "Style: use early-return to clarify code"

I still need to check the call of updatemons

PR #475, opened 2023-09-05
Fix undefined behaviour in xytonode

NikitaIvanovV:
Sometimes `pnode->parent` equals NULL.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -2638,8 +2638,11 @@ xytonode(double x, double y, struct wlr_surface **psurface,
 			surface = wlr_scene_surface_from_buffer(
 					wlr_scene_buffer_from_node(node))->surface;
 		/* Walk the tree to find a node that knows the client */
-		for (pnode = node; pnode && !c; pnode = &pnode->parent->node)
+		for (pnode = node; pnode && !c; pnode = &pnode->parent->node) {
 			c = pnode->data;
+			if (!pnode->parent)
+				break;
+		}
 		if (c && c->type == LayerShell) {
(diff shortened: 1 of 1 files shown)


sevz17:
In what case?, wlroots does not allow NULL parents.

NikitaIvanovV:
I had some crashes which were I could not properly reproduce. So I compiled with `-fsanitizer=undefined,address` and it exposed that something weird is happening in that for loop. After adding that check, the crashes stopped happenning.

Maybe there is a bug in wlroots that causes parent become NULL?

sevz17:
The only reason I think this would happen is if you system was on high load.

Otherwise please attach a full coredump with wlroots' debug symbols

dwl.c
```diff
@@ -2638,8 +2638,11 @@ xytonode(double x, double y, struct wlr_surface **psurface,
 			surface = wlr_scene_surface_from_buffer(
 					wlr_scene_buffer_from_node(node))->surface;
 		/* Walk the tree to find a node that knows the client */
-		for (pnode = node; pnode && !c; pnode = &pnode->parent->node)
+		for (pnode = node; pnode && !c; pnode = &pnode->parent->node) {
```
sevz17, line 2641: @NikitaIvanovV, after a second look, does this work?
```suggestion
		for (pnode = node; pnode && !c && pnode->parent; pnode = &pnode->parent->node) {
```

NikitaIvanovV:
I tried to undo the change but could not reproduce the crash in order to send the coredump, closing the issue for now.

Issue #476, opened 2023-09-06
clicking using trackpoint is a pain on some screens


pony-montana:
## Info
dwl version: here https://github.com/pony-montana/simple-dwl is based on lastest master commit + some patches
wlroots version: 0.16.2
## Description

Using trackpoint and clicking mouse-integrated-buttons (left button in particular) on my thinkpad x200 result in a painful experience on some screens. 

The problem is: when I click, the screen registers some micro-movements from trackpoint input; so clicking doesnt produce effects (or, in some screens, it produce the effect of "keep-clicked-and-drag").

Only if I click the left button when I'm not touching the trackpoint all goes fine. But this result in a pretty slow user experience, not the way a trackpoint is intended to be used.

In particular, I found this problem in some qt-screens (navigation menus in electrum bitcoin wallet) and in firefox options.
For example, clicking on web-content and on the main firefox interface works fine. The problem comes only on some screen, when I click on menus that sovrappose to the main screen.
This problem is not present when I click on dwl-bar for example.

sevz17:
Hmm, maybe [related](https://wayland.freedesktop.org/libinput/doc/1.24.0/touchpad-jumping-cursors.html)?

sevz17:
Can you reproduce in sway (or in another non-minimal compositor)?

pony-montana:
Using sway I don't have this issue (on the same machine, with all the same libraries and same environment).
So I think that there are problem with compositing. On screens where I have problems I feel like it "reacts to trackpoint inputs de-focusing from the thing that is on-focus in that moment" where it would just move the cursor on the screen and nothing else.

pony-montana:
The problem is also present on plain dwl last main commit and plain dwl 4.0.

pony-montana:
Found a dirty solution, made a the pull request #478 
Hope this will help!

sevz17:
Could you test in wlroots-next?, I suspect this happens because a hack I did.

pony-montana:
Tested wlroots-next branch with wlroots compiled at last commit and still the issue is present. 

I isolated some lines in the code (from main branch) that when are changed the issue is solved (but other non-intended behaviour happens):

1)
in dwl.c, if I remove the lines number 1691 and 1692 in the motionnotify function like that:

dwl.c
```c
if (cursor_mode == CurPressed && !seat->drag) {
		if ((type = toplevel_from_wlr_surface(
				 seat->pointer_state.focused_surface, &w, &l)) >= 0) {
[...]

PR #478, opened 2023-09-09
dirty solution to issue #476

pony-montana:
Hope this PR help figuring out whats the root cause of the problem.


pony-montana:
okay, studied a bit what break was intende to do here.
Close that pull request, will come later when I will be able to understand better what I'm doing.

PR #481, opened 2023-09-11
Fixed Typo

two-horned:
The backtick symbol was at the wrong position.
If you wanted to compile with XWayland support 
(by commenting according lines out in config.mk),
you'd run into an error.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,7 @@ DWLDEVCFLAGS = -pedantic -Wall -Wextra -Wdeclaration-after-statement -Wno-unused
 # CFLAGS / LDFLAGS
 PKGS      = wlroots wayland-server xkbcommon libinput $(XLIBS)
 DWLCFLAGS = `$(PKG_CONFIG) --cflags $(PKGS)` $(DWLCPPFLAGS) $(DWLDEVCFLAGS) $(CFLAGS)
-LDLIBS    = `$(PKG_CONFIG) --libs $(PKGS)` $(LIBS)
+LDLIBS    = `$(PKG_CONFIG) --libs $(PKGS) $(LIBS)`
 
 all: dwl
 dwl: dwl.o util.o


sevz17:
No, you wouldn't.
Check carefully, what you uncomment in `config.mk` is `XLIBS`, which is appended to `$(PKGS)`.

`LIBS` is just to pass raw libs to the linker, eg. another way to enable Xwayland support is append -DWAYLAND to your CPPFLAGS, and export `LIBS="-lxcb -lxcb-icccm`

two-horned:
contents of my config.mk (when compiling with xwayland support):
```
_VERSION = 0.4
VERSION  = `git describe --tags --dirty 2>/dev/null || echo $(_VERSION)`

PKG_CONFIG = pkg-config

# paths
PREFIX = ~/.local
MANDIR = ~/.local/man

#XWAYLAND =
#LIBS =
# Uncomment to build XWayland support
XWAYLAND = -DXWAYLAND
LIBS = xcb xcb-icccm
```

Any changes unrelated to the Makefile or config.mk shouldn't affect the following compilation error.
The error I got:
```
/usr/bin/ld: cannot find xcb: No such file or directory
/usr/bin/ld: cannot find xcb-icccm: No such file or directory
collect2: error: ld returned 1 exit status
make: *** [Makefile:18: dwl] Error 1
```

The changes mentioned in the PR have fixed the error.
Do with this information what you want.

You should be able to reproduce the same error by cloning the main branch
and compiling it with the above config.mk.

Again, do with this information what you want.

fauxmight:
@two-horned It appears you do not have libxcb installed.

sevz17:
It's not that he doesn't have it, but...

@two-horned somehow your config.mk has `LIBS` instead of `XLIBS`, that's why it fails to compile, rename it to `XLIBS` and it should work.

two-horned:
> It's not that he doesn't have it, but...
> 
> @two-horned somehow your config.mk has `LIBS` instead of `XLIBS`, that's why it fails to compile, rename it to `XLIBS` and it should work.

Thank you for your patience and sorry for the trouble.
I got a bit too frustrated with the compilation issue so I thought there was a mistake with the Makefile, but I was wrong.

You're right about the config.mk file.

PR #485, opened 2023-09-21
Fix tiling order for X11 clients

fbushstone:
Currently, Wayland clients are tiled in the correct order, but clients running under XWayland are not. To see this in action, spawn (for example) st, chromium and st and observe that the visible tiling order is st, st, chromium.

Applying this patch fixes this behavior.

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -2679,8 +2679,6 @@ zoom(const Arg *arg)
 	 * front */
 	if (!sel)
 		sel = c;
-	wl_list_remove(&sel->link);
-	wl_list_insert(&clients, &sel->link);
 
 	focusclient(sel, 1);
 	arrange(selmon);


dwl.c
```diff
@@ -2679,8 +2679,6 @@ zoom(const Arg *arg)
 	 * front */
 	if (!sel)
 		sel = c;
-	wl_list_remove(&sel->link);
-	wl_list_insert(&clients, &sel->link);
 
 	focusclient(sel, 1);
```
fbushstone, line 2683: It looks like the focusclient here causes some weird behavior when closing X11 windows (i.e. the next window in the fstack is not necessarily selected).

fbushstone:
Whoops, looks like I forgot to unbind XK_Return from `zoom` in config.h. Not sure how I missed that, meaning this is not an issue.

Issue #487, opened 2023-09-25
Make sure that background does not spawn in front of fullscreen clients


fbushstone:
I think it would be a good idea to document what exactly `fullscreen_bg` does. I'd spent the past few hours trying to figure out why SDL apps such as Quake and Xonotic were hidden behind a gray screen while the glx version of Xonotic works fine. Like the config file says, changing the alpha value to `0.0` solves this issue. Maybe something such as: "If fullscreen SDL apps are covered by a solid color, change this to 0.0."  Although, maybe I should read the XDG protocol :smile:.

fauxmight:
@fbushstone This sounds reasonable. I am not among the dev's for dwl, but your suggestion might be quicker implemented with a pull request.

fbushstone:
Yes, although I suppose this issue could be a discussion on whether to remove this option as well (making it internal only). No other Wayland compositor I tested had this feature (too intentional to be a bug) and I'm not sure what it's used for.

sevz17:
> I think it would be a good idea to document what exactly `fullscreen_bg` does. I'd spent the past few hours trying to figure out why SDL apps such as Quake and Xonotic were hidden behind a gray screen while the glx version of Xonotic works fine. Like the config file says, changing the alpha value to `0.0` solves this issue. Maybe something such as: "If fullscreen SDL apps are covered by a solid color, change this to 0.0." Although, maybe I should read the XDG protocol 😄.

Err... the behavior you mentioned is a bug, the gray screen should be behind fullscreen apps, not the other way around.

Respecting the "old behavior":
the protocol [says](https://gitlab.freedesktop.org/wayland/wayland-protocols/-/blob/main/stable/xdg-shell/xdg-shell.xml#L1048-1057):
```
	If the surface doesn't cover the whole output, the compositor will
	position the surface in the center of the output and compensate with
	with border fill covering the rest of the output. The content of the
	border fill is undefined, but should be assumed to be in some way that
	attempts to blend into the surrounding area (e.g. solid black).

	If the fullscreened surface is not opaque, the compositor must make
	sure that other screen content not part of the same surface tree (made
	up of subsurfaces, popups or similarly coupled surfaces) are not
	visible below the fullscreened surface.

```
dwl ignored that for a long time and rendered other apps behind fullscreen windows, that's why setting the alpha to 0 restores the old behavior (it "renders" an invisible solid color, which is the same to render nothing)
and it's an option because it's pretty useful with transparent terminals

[...]

PR #488, opened 2023-09-25
Draw fullscreen_bg on the correct layer

fbushstone:
Closes [Issue 487](https://github.com/djpohly/dwl/issues/487).
dwl would draw the background on the wrong layer, obscuring SDL apps such as Quake and making certain terminals totally transparent except for the text (adjusting alpha on fullscreen_bg would not have helped in this case).

### DIFF
--- a/dwl.c
+++ b/dwl.c
@@ -953,7 +953,7 @@ createmon(struct wl_listener *listener, void *data)
 	 *
 	 */
 	/* updatemons() will resize and set correct position */
-	m->fullscreen_bg = wlr_scene_rect_create(layers[LyrFS], 0, 0, fullscreen_bg);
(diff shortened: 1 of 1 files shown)


sevz17:
Hm
I don't think this will work, this could render the background behind the wallpaper (I didn't test).
And I think it's better to avoid fullscreen windows being in other layer than `LyrFS` and make sure that `fullscreen_bg` is always at the bottom of `LyrFS`

fbushstone:
@sevz17 Would it make any sense to add another layer for just the fullscreen backgroun (e.g. LyrFSBg)?

fbushstone:
On second inspection, I shouldn't have pushed that change. It seems like certain clients request fullscreen mode, are actually fullscreen, but then dwl doesn't consider them fullscreen internally for whatever reason.

fbushstone:
Poor solution on my part, see my updated pull request

PR #491, opened 2023-09-29
Make sure that fullscreen clients are on correct layer

fbushstone:
Closes [Issue 487](https://github.com/djpohly/dwl/issues/487).
Should actually fix the issue rather than just bandaid over it like by earlier [PR](https://github.com/djpohly/dwl/pull/488).

Clients were that had fullscreen enabled were reparented to the floating layer rather than the fullscreen layer. This should allow clients that have fullscreen enabled to be set to the correct layer.


fbushstone:
This change could also be implemented by changing `layers[c->isfloating ? LyrFloat : LyrTile]` to `layers[c->isfullscreen ? LyrFS : c->isfloating ? LyrFloat : LyrTile] in line 2031.

sevz17:
Thanks!

Issue #427, opened 2023-05-31
Move chvt into core DWL instead of the config.h


godalming123:
One of my first experiences with DWL was managing to break virtual terminal switching and I think it should be mined into dwl.c instead.

PalanixYT:
I believe this wouldn't be worth the extra sloc. @sevz17 thoughts?

PR #493, opened 2023-10-06
add a comment about chvt keybindings

sevz17:
Closes: https://github.com/djpohly/dwl/issues/427

### DIFF
--- a/config.def.h
+++ b/config.def.h
@@ -154,6 +154,9 @@ static const Key keys[] = {
 
 	/* Ctrl-Alt-Backspace and Ctrl-Alt-Fx used to be handled by X server */
 	{ WLR_MODIFIER_CTRL|WLR_MODIFIER_ALT,XKB_KEY_Terminate_Server, quit, {0} },
+	/* Ctrl-Alt-Fx is used to switch to another VT, if you don't know what a VT is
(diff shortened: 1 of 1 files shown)


sevz17:
Probably there is a better way to let the users know they are important, I am open to suggestions.

pm4rcin:
I think it's fine to merge because IMO people who use standalone WMs know what a VT is. It's not like many new to linux users start with WMs rather than DE. I wonder if any linux user started with WMs.

config.def.h
```diff
@@ -146,6 +146,9 @@ static const Key keys[] = {
 
 	/* Ctrl-Alt-Backspace and Ctrl-Alt-Fx used to be handled by X server */
 	{ WLR_MODIFIER_CTRL|WLR_MODIFIER_ALT,XKB_KEY_Terminate_Server, quit, {0} },
+	/* Ctrl-Alt-Fx is used to switch to another VT, if you don't know what a VT is
+	 * Do not remove them
```
pm4rcin, line 150: Not sure if capitalization of "Do" is needed. Also I don't get what does the * in that line do? I guess it's some convention or smth.

sevz17: Ups, it was a typo

PR #494, opened 2023-10-06
color the border color of urgent clients red

sevz17:
iirc this is the same behavior of dwm


sevz17:
I don't quite like the phrasing of the commit messages (especially the second), if you have any suggestions please tell me!

PR #496, opened 2023-10-10
check client is mapped before setting border color

bencollerson:
For some reason brave configured for as a wayland client triggers this code on startup and segfaults.

Checking if the client is mapped fixes this, like with the previous fix for urgent border colour.


sevz17:
Thank you!

